PR #73, opened 2014-04-21
Allow grouping with m_ldap_authentication

KlaasT:
This would allow people to group their nicknames even when m_ldap_authentication is in use. The grouped nicks won't be added as a new entry in the LDAP directory.


Adam-:
If we ever split nicks from accounts (2.2?) we could do this, but currently it's not really an option.

PR #90, opened 2014-10-19
Support for cloning ChanServ levels between channels

Genesis2001:
I needed a way to clone levels between two channels, so I modified cs_clone and a core header (and associated source) file to add this functionality.

Modification compiles and runs smoothly on my network.

Sorry for the bad commit name in the diff. I tried to edit it, but it didn't take apparently.


Genesis2001:
pooke

Adam-:
I have applied this myself in 310e95a92e64a9546247fe57d4643328fa4ebe9b

PR #94, opened 2014-12-11
Fix issue 1627.

webczat:
It required that the FHost message was rewritten for the insp20 protocol module instead of relying on inspircd12's version.
The insp20 version internally removes +x mode on host change.

### DIFF
--- a/modules/protocol/inspircd20.cpp
+++ b/modules/protocol/inspircd20.cpp
@@ -782,6 +782,19 @@ struct IRCDMessageEncap : IRCDMessage
 	}
 };
 
+struct IRCDMessageFHost : IRCDMessage
+{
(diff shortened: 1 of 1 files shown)


Adam-:
What problems were you seeing prior to this commit? This is causing bug #1635

webczat:
I am not quite sure if I remember it right, but the issue was probably that when inspircd accepted a host change from external sources, it automatically assumed +x being removed. But the mode change was not synced, and this caused anope to leave +x set even when the rest of the network saw it unset. The only thing that was sent to every server was a host change itself and all servers would internally -x the user.

Adam-:
I found it, this PR is for bug #1627. I'm going to reorder this so -x is set first before setting the vhost to fix #1635.

webczat:
Well, and it seems I found why that happened (in code)... internally removing +x cleared the vhost, right? so what the fhost handler did was nullified few lines later...

Adam-:
Yes.

PR #98, opened 2015-01-03
cs_entrymsg to use SET permission, not isFounder

flotwig:
Makes the ability to control ENTRYMSG's assignable instead of restricted to only the channel founder.

### DIFF
--- a/modules/commands/cs_entrymsg.cpp
+++ b/modules/commands/cs_entrymsg.cpp
@@ -143,7 +143,7 @@ class CommandEntryMessage : public Command
 		else
 		{
 			(*messages)->push_back(new EntryMsgImpl(ci, source.GetNick(), message));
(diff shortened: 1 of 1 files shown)


flotwig:
Okay, I'm a little lost here. I don't know C++ to fix this, but now it says the list is full whenever trying to add a new ENTRYMSG. Someone want to point out the problem?

Adam-:
You have entrymsg:maxentries set to 0.

flotwig:
It's set to 5, ENTRYMSG worked (only for founder) before this change. That's why it's perplexing.

flotwig:
Patch works correctly, it was an error on our side for having the module renamed. Everything is a-ok.

Adam-:
Can you squash this?

flotwig:
There you go.

PR #99, opened 2015-02-05
Added kickcheck override for users with protected privs

H7-25:
I think should be useful having this override for users with protected privs (Can not be kicked from channels by Services)


Adam-:
Probably just u->isProtected()

H7-25:
i don't see isProtected() in user class ..

Adam-:
I meant IsProtected

Adam-:
Squash this

H7-25:
Squashed :+1:

Adam-:
I don't know why you want this, this is redundant due to https://github.com/anope/anope/blob/2.0/src/channels.cpp#L896.

PR #108, opened 2015-02-16
Use gender neutral pronoun

Mikaela:
I am not sure if I got all instances, I was going through GitHub search.
Please tell me if I missed something and where.


Mikaela:
I also didn't update language files as they seemed to be already outdated.

PR #110, opened 2015-03-02
win32.rc error RC2127

Zageron:
CMake broke as of this commit, the win32.rc was reporting incorrect versions with error RC2127 : version WORDs separated by commas expected. This is definitely not a solution, but the easiest way to report the issue. As of now building on Windows is blocked.

The area in question reports:
VER_ANOPE VERSIONINFO
 FILEVERSION ,,,0
 PRODUCTVERSION ,,,0

Rather than
VER_ANOPE VERSIONINFO
 FILEVERSION 2,0,2,0
 PRODUCTVERSION 2,0,2,0


Zageron:
So given the failure of the build, I'd say a different approach will need to be taken to solve this. I'm not familiar with CMake, so I do not know how to resolve it. Cheers,

Adam-:
This commit is okay, the Windows specific build system code has not been updated to cope with this new change is all.

Zageron:
Alright, good to know. I'll keep my eye open for the build system updates, thank you for the speedy response. :)

Adam-:
This is fixed now.

PR #116, opened 2015-04-01
Fix error and make file comply with PSR-2 coding standard

aFreshMelon:


aFreshMelon:
Will split up into separate commits.

PR #118, opened 2015-04-05
Remove max password length from Anope

sebastianbarfurth:
I'm not sure why this was implemented, but a max password length is considered bad practice in any case since it is irrelevant for hashed passwords anyways.

### DIFF
--- a/data/nickserv.example.conf
+++ b/data/nickserv.example.conf
@@ -203,13 +203,6 @@ module
 	 * If set, Services do not allow ownership of nick names, only ownership of accounts.
 	 */
 	nonicknameownership = no
-
-	/*
-	 * The maximum length of passwords
-	 *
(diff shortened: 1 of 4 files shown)


Adam-:
Passwords aren't always hashed. And it's possible to send a password that is so long the IRCd silently truncates it, so you don't know what password was really set. Even if it does reply with the password you set (which it can depending on crypto settings), that could be truncated even more so, depending on lengths of the various nicknames/hostnames etc in the message.

aFreshMelon:
This would clearly be an endusers fault, since passwords longer than the IRCd message limit are very uncommon and quite unreasonable. It is just fairly uncommon to have limits on passwords.

Where are passwords not hashed, by the way?

Adam-:
enc_none?

PR #119, opened 2015-04-30
irc2sql: handle away messages.

Jyzee:


Adam-:
Shouldn't there be some sort of schema change in here somewhere?

Jyzee:
These fields are already there. See https://github.com/anope/anope/blob/2.0/modules/extra/stats/irc2sql/tables.cpp#L120

PR #120, opened 2015-05-11
Fix db_flatfile to read the "nobackokay" setting

Alef-Burzmali:
The code only looks for 'nobackupok' and 'nobackupokay' was ignored.


Adam-:
Fix the code to read the "nobackupokay" instead.

I think we should just totally remove this option though it's silly, maybe in 2.2

SaberUK:
You probably want to fix the pull request title too.

Alef-Burzmali:
Sure :)

PR #121, opened 2015-05-23
Various improvements to the build system.

SaberUK:
## Remove an old Autotools limitation.

This is not needed anymore as support for Autotools was removed. Removing it also makes it easier for package generation as you don't need to manually create a build directory.

## Don't overwrite CMAKE_INSTALL_PREFIX if it has already been set.

This helps as packaging systems often have a specific set of arguments they pass to CMake which includes CMAKE_INSTALL_PREFIX.

## Fix matching the Apple Clang variant on CMake 3.0+

Since CMake 3.0 the Apple variant of Clang has been called AppleClang.

## Remove support for run-cc.

As discussed with @Adam- in #inspircd-dev this tool does not really provide any benefits in 2015. It does however cause a rise in the amount of time it takes to build stuff. It has also recently been removed from InspIRCd in commit  https://github.com/inspircd/inspircd/commit/01dee272ca360fe8cd227583061ec5a11c2dadc1.


Adam-:
Thanks

PR #130, opened 2015-07-16
Read All Memos

AlphaT3ch:
Allows user to do "/msg MemoServ READ ALL" as requested in [Bug #0001221](http://bugs.anope.org/view.php?id=1221)

### DIFF
--- a/modules/commands/ms_read.cpp
+++ b/modules/commands/ms_read.cpp
@@ -169,7 +169,7 @@ class CommandMSRead : public Command
 			}
 			else if (numstr.equals_ci("ALL"))
 			{
-				for (i = mi->memos->size(); -1 < i; --i)
+				for (i = 0; i < mi->memos->size(); i++)
 				{
 					MemoListCallback::DoRead(source, mi, ci, i);
 				}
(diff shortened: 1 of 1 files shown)


del6597:
Why does it read the memos in reverse order?  NEW reads them incrementally, so why is this different?
It may show you the newest memos first, but in the case of a user sending a large message across many memos, then they will appear out of order.

AlphaT3ch:
Thanks for the feedback, @del6597, I've fixed the error and added it to the fork :+1:

JeDa:
Looks nice :+1:

modules/commands/ms_read.cpp
```diff
@@ -167,6 +167,13 @@ class CommandMSRead : public Command
 				for (i = 0, end = mi->memos->size() - 1; i < end; ++i);
 				MemoListCallback::DoRead(source, mi, ci, i);
 			}
+			else if (numstr.equals_ci("ALL"))
+			{
+				for (i = 0; i < mi->memos->size(); i++)
```
del6597: It may be convention to use ```++i``` here, instead of ```i++```.  I guess that's up to @Adam- though.

Adam-:
I guess this is okay. Squash the commits.

AlphaT3ch:
Opened as PR #135

Adam-:
You know you can just force push the branch instead of creating an entire new pr right. That just creates useless spam.

AlphaT3ch:
@Adam- Oh whoops, my bad, sorry

PR #131, opened 2015-07-20
Did it send? BotServ Enhancement

AlphaT3ch:
Some may say this is pointless, but I think it adds a good touch. When using "/bs say" or "/bs action" it will tell the user via notice if the message has sent instead of just staring into blank nothing. Some newer users may think the services are broken when they're just not checking the right tab.

->BotServ<- say #Test This is a test!
<ChanServ> This is a test!
-BotServ- Sent message to channel

### DIFF
--- a/modules/commands/bs_control.cpp
+++ b/modules/commands/bs_control.cpp
@@ -56,6 +56,7 @@ class CommandBSSay : public Command
 		}
 
 		IRCD->SendPrivmsg(*ci->bi, ci->name, "%s", text.c_str());
+		source.Reply("Sent message to channel", ci->name);
 		ci->bi->lastmsg = Anope::CurTime;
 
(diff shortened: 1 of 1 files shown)


Zoddo:
I think it will flood the sender than anything else.

The user know in what channel it sends the message.

Adam-:
I'd be okay with this only if the user doing the sending isn't in the channel.

Zoddo:
Or allow to disable this notice.

AlphaT3ch:
@Adam- That's possible? Doing "!help say" says that there is no help for that command (therefor not a command)

del6597:
@AlphaT3ch there is help for the SAY command.  You can get the help by messaging BotServ help say.
Syntax: SAY _channel_ _text_
"Makes the bot say the specified text on the specified channel."

AlphaT3ch:
@del6567 And?

PR #132, opened 2015-07-22
Add support for sending messages in the context of a channel.

SaberUK:
Such messages appear in the channel window but are only sent to the user. This prevents the problem of getting spammed with private messages when connecting to a server and joining a ton of channels.

Presently only support for InspIRCd is implemented as that is the only protocol which I am familiar with. Servers which do not support this will fall back to the old method.

### DIFF
--- a/data/example.conf
+++ b/data/example.conf
@@ -262,6 +262,15 @@ module
 	name = "inspircd20"
 
 	/*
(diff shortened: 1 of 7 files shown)


del6597:
Nice work @SaberUK.
I'm interested in using this feature on my network, so thanks for writing this so quickly.

SaberUK:
Closing this as it seems there is no interest in it.

PR #133, opened 2015-07-28
os_akill: Sanity check AKILLing before modifing the akill list

staticfox:
When a user issues an AKILL, CommandOSAkill::DoAdd checks against XLineManager::CanAdd before the mask can completely be dry tested. This creates the issue that if the mask fails dry tests past XLineManager::CanAdd, XLineManager has already removed any matching masks while DoAdd still fails the final result (entry). This patch reorders the dry checks and places the call to XLineManager::CanAdd as the final check before proceeding to add. That way, if it fails a dry check prior to CanAdd, there is no risk of modifying the current AKILL list.


Adam-:
Why not only move the canadd check? I don't like constructing xlines here with these checks removed

blindirc:
Apologies for the delay, I've amended your requested changes.

Adam-:
Merged

PR #134, opened 2015-08-13
Have CMake first check for libraries in any additional locations

retsofaj:
Followed by the standard locations. This allows (e.g.) using the Homebrew-installed OpenSSL instead of the system version on OS X.


cmake/Anope.cmake
```diff
@@ -449,7 +449,8 @@ macro(calculate_libraries SRC SRC_LDFLAGS EXTRA_DEPENDS)
       if(DEFAULT_LIBRARY_DIRS OR WSDK_PATH OR DEFINED $ENV{VCINSTALLDIR})
         find_library(FOUND_${LIBRARY}_LIBRARY NAMES ${LIBRARY} PATHS ${DEFAULT_LIBRARY_DIRS} ${WSDK_PATH}/lib $ENV{VCINSTALLDIR}/lib ${EXTRA_INCLUDE} ${EXTRA_LIBS})
```
Adam-: Maybe this should do that too

retsofaj: I'm don't have a setup to test on Windows (which I'm assuming that code path is for) which is why I restricted my changes to only the path that my system was following. The additional changes would make sense, but absent testing I figured conservative was better.

PR #137, opened 2015-10-25
Send client introduction+valid ip format

miwob:


Adam-:
Cherrypicked as this pr looks messed up.

PR #151, opened 2016-02-05
Updated Travis build process for Anope 2.1

KeiroD:
It took a fair amount of time to get a working process for Travis CI to build Anope 2.1 successfully.

Mostly because I was not successfully getting gcc to install. But now we're pretty good to go. If nothing else, it's good to have a method for checking your code works... after all, you guys have it for Anope 2.x.


KeiroD:
Missed one item, will do another PR, sorry about this.

PR #154, opened 2016-02-10
Fix a low risk privilege escalation exploit on InspIRCd with m_autoop.

SaberUK:
It isn't as serious as it sounds as you need access to change channel modes via ChanServ to perform the escalation which by default requires channel operator privileges. However, if you have those privileges and the server has status modes higher than op (i.e. +Yyao) enabled then you can set +w masks to give people those any of those status modes.

Workaround for this bug is to unload m_autoop on InspIRCd. You don't need it anyway if you're using Anope.


Adam-:
You can set chanserv:nomlock = "w" too,

PR #157, opened 2016-03-07
Prevent using enc_none as the main encryption module.

SaberUK:
Plain text passwords are a terrible idea at best and are illegal at worst. Lets not support them.


SketchNI:
While I do agree with you, @SaberUK, people migrating from 1.x might have plaintext passwords and need enc_none to migrate to 2.x before applying encryption. (If I'm thinking of the process correctly)

SaberUK:
@SketchNI I changed it to refuse to load as the primary encryption provider. It can still be loaded as a secondary provider which is what is used if someone is upgrading.

SketchNI:
Ah, my bad! I misunderstood :)

DukePyrolator:
This patch also obsoletes ns_getpass, because ns_getpass requires enc_none to be loaded as primary encryption module.

SaberUK:
@DukePyrolator I'll remove that then

Robby-:
Also needs `nickserv/getpass` to be removed from [here](https://github.com/anope/anope/blob/2.0/data/example.conf#L740) and [here](https://github.com/anope/anope/blob/2.0/data/example.conf#L796).

SaberUK:
I can also remove `Anope::Decrypt` too. What do y'all think?

bonnedav:
I think that you should mention this in the config files.

SaberUK:
@bonnedav This was already done a while back.

https://github.com/anope/anope/blob/master/data/anope.example.conf#L1164-L1170

bonnedav:
I mean you should note that enc_none will no longer work as the primary encryption module.

del6597:
I think the suggestion is to update this.
` * With encryption/none, passwords will be stored in plain text, allowing for passwords
 * to be recovered later but it isn't secure and therefore is not recommended.`

SaberUK:
Is that better?

bonnedav:
Yes it is, thank you. Although this suggests that enc_md5, enc_sha1, and enc_old also may not be used as the primary encryption module. Is this the case?

SaberUK:
Yes, that is the case. The other modules were deprecated in an earlier commit. The only modules you can use as the primary in 2.1 are bcrypt and sha256.

bonnedav:
Ok then.

SaberUK:
:tada:

PR #158, opened 2016-03-16
protocol/charybdis: add support for MECHLIST

kaniini:


Adam-:
If someone else wants to re-implement this, test it, and submit it, I will accept it. But I will not accept any pull requests from @kaniini.

PR #161, opened 2016-04-10
Send vident/vhost in SVSLOGIN with SASL

Zoddo:
Only charybdis seems to support that for the moment.


SaberUK:
Why can't you just call `GetVhostIdent()` and `GetVhostHost()` from the Charybdis protocol module? That way you don't need to update the entire protocol API just for one IRCd.

PR #167, opened 2016-04-23
A fix for a 404 error when using the DoNotice method.

SketchNI:


Adam-:
Thanks, fixed in 4c4cc0ded7889cc65ee7faafd0e8b9c33421fd8d.

PR #168, opened 2016-05-19
Update chanstats.example.conf

SketchNI:
Added in extra (optional) smileys.

### DIFF
--- a/data/chanstats.example.conf
+++ b/data/chanstats.example.conf
@@ -20,9 +20,22 @@ module
 	 */
 	prefix = "anope_"
 
+	/*
+	 * Default smileys. These come with Anope.
+	 * Comment these out before uncommeting the extended smileys.
+	 */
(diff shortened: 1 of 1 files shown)


SaberUK:
Is there any reason you didn't just add it to the existing smiley list? If someone doesn't want them then they can be removed easily.

SketchNI:
It'll save time (seconds, really) for people that don't want them.

data/chanstats.example.conf
```diff
@@ -20,9 +20,22 @@ module
 	 */
 	prefix = "anope_"
 
+	/*
+	 * Default smileys. These come with Anope.
+	 * Comment these out before uncommeting the extended smileys.
```
GLolol: uncomme**n**ting, not uncommeting

Adam-:
This is ridiculous. Add them to the default and perhaps.

Adam-:
Reopen if you want to change this.

PR #171, opened 2016-07-22
charybdis.cpp: Properly support RESVs

Techman-:
Properly support RESV and UNRESV in Charybdis, and improve RESV setting for pseudoclients when bursting.


Zoddo:
Tested on Charybdis 3.5.2, works well: RESVs (both for nicks and channels) are set/unset correctly.

PR #174, opened 2016-08-12
InspIRCd: Send CHGIDENT/CHGHOST on SVSLOGIN

Techman-:


modules/protocol/inspircd12.cpp
```diff
@@ -410,6 +410,12 @@ class InspIRCd12Proto : public IRCDProto
 	{
 		UplinkSocket::Message(Me) << "METADATA " << uid << " accountname :" << acc;
 
+		if (Servers::Capab.count("CHGIDENT") && !vident.empty())
+			UplinkSocket::Message(Me) << "CHGIDENT " << uid << " " << vident;
```
attilamolnar: This must use ENCAP to get the message forwarded to the correct server if `uid` is a not yet registered user.

modules/protocol/inspircd12.cpp
```diff
@@ -410,6 +410,12 @@ class InspIRCd12Proto : public IRCDProto
...
+
+		if (Servers::Capab.count("CHGHOST") && !vhost.empty())
+			UplinkSocket::Message(Me) << "CHGHOST " << uid << " " << vhost;
```
attilamolnar: Same as above.

PR #178, opened 2016-09-04
ratbox: fix RESVs

Techman-:
Also, enable channel support because it wasn't specified already and point charybdis to these methods instead.


Zoddo:
Ok, nickname and channels RESV/SQLine work well :)

PR #181, opened 2016-11-11
Update info.cpp

Gavin-X:
fixed a typo

### DIFF
--- a/modules/operserv/info.cpp
+++ b/modules/operserv/info.cpp
@@ -236,7 +236,7 @@ class CommandOSInfo : public Command
 
 	bool OnHelp(CommandSource &source, const Anope::string &subcommand) override
 	{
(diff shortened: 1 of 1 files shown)


modules/operserv/info.cpp
```diff
@@ -236,7 +236,7 @@ class CommandOSInfo : public Command
 
 	bool OnHelp(CommandSource &source, const Anope::string &subcommand) override
 	{
-		source.Reply(_("Add or delete oper information for a given accpint or channel. This informaition will show to opers in the respective info command for the account or channel."));
+		source.Reply(_("Add or delete oper information for a given account or channel. This informaition will show to opers in the respective info command for the account or channel."));
```
SaberUK: Information is misspelled too.

PR #184, opened 2017-01-22
SASL: disallow unconfirmed accounts from successfully authenticating

Techman-:
The reason for this is that unconfirmed accounts can connect to a network and authenticate using SASL in order to get their account name set, which allows the client to then bypass +R and other popular registration-limited channel modes on many popular IRCds. Clients that identify via NickServ do not have their account names set if they are unconfirmed, and thus are not affected by this.


Adam-:
Thanks

PR #190, opened 2017-05-27
Removed obsolete switch cases and combined some if statements

key2peace:


key2peace:
Removing coz this way the PR sucks

PR #191, opened 2017-05-28
Removed obsolete switch cases and combined some if statements

key2peace:


Lord255:
mhm.
drop support for inspir v1.2?
https://www.ircstats.org/servers/inspircd
if so, then this issue can be closed as well. (or dropping support doesnt even matter. 2017!)

PR #192, opened 2017-06-25
Add RobustIRC module

stapelberg:
See https://robustirc.net/ for details.

The module is based on the UnrealIRCD module (minus a bunch of features which
aren’t included in RobustIRC) and works with RobustIRC v0.5 and newer.

### DIFF
--- a/modules/protocol/robustirc.cpp
+++ b/modules/protocol/robustirc.cpp
@@ -1,0 +1,537 @@
+/* RobustIRC
+ *
+ * (C) 2003-2014 Anope Team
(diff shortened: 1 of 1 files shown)


SaberUK:
>not supported

>not supported

>not supported

Is there really any point in supporting an extremely niche pseudo-IRCd that doesn't support half of the features required?

stapelberg:
I think so, otherwise I wouldn’t have sent the pull request :).

Regarding features: the existing user-base is happy with the current functionality the services provide. As requests come in, we can implement more features — one has to start somewhere.

PR #193, opened 2017-07-24
Don't require a user to GROUP nicknames

kaecyra:
Attempting to patch https://bugs.anope.org/view.php?id=1704

Prior to this patch, XMLRPC requests to nickserv/group would report success but actually fail, because NSGroupRequest::OnSuccess explicitly checked for source.GetUser, which is never present in an XML RPC call.

~~This patch attempts to bypass this problem by diverting userless calls to a secondary codepath that does not attempt to adjust a user object.~~

### DIFF
--- a/modules/commands/ns_group.cpp
+++ b/modules/commands/ns_group.cpp
@@ -24,7 +24,10 @@ class NSGroupRequest : public IdentifyRequest
 
 	void OnSuccess() anope_override
 	{
-		if (!source.GetUser() || source.GetUser()->nick != nick || !target || !target->nc)
+		if (!source.GetUser())
+			return this->GroupUserless();
+
+		if (source.GetUser()->nick != nick || !target || !target->nc)
 			return;
 
 		User *u = source.GetUser();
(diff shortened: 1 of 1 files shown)


kaecyra:
This has been tested both with XMLRPC and with normal IRC grouping/ungrouping and it works. I also tested that grouping is NOT done when the password is incorrect ;)

Adam-:
Can you stead https://github.com/Adam-/anope/commit/0b7b6d9d6d55fc0398fb330e95c43e1c34c6ae2d? this is more similar to how registration works and I want to keep it consistent.

kaecyra:
I really want to keep it as similar to the rest of Anope as possible, so I am keen to make any changes you recommend. However, there is no `source.GetUser()` available when this is called from XMLRPC. That's why I did what I did. 

from m_xmlrpc_main:

```
CommandSource source(user, NULL, na ? *na->nc : NULL, &reply, bi);
Command::Run(source, command);
```

That first `NULL` is where a User would get passed in, as far as I understand it...

If you have a suggestion for how to get an instance of `User` for the calling user, then we don't even need a second codepath. But I couldn't see how to do that, especially since the command can be called by XMLRPC even when the user is not online.

Does that make sense?

kaecyra:
Sure :)

Adam-:
Sorry my previous comment was missing a word or two. I mean can you test https://github.com/Adam-/anope/commit/0b7b6d9d6d55fc0398fb330e95c43e1c34c6ae2d? Thanks

kaecyra:
Yeah. I refactored this PR into something a lot like what you linked, and tested it. Works. We are using it in production now.

Adam-:
Okay, merged.

PR #194, opened 2017-07-24
Detect and handle unclosed tags in XMLRPC content.

kaecyra:
Fixes #1703 by detecting unclosed tags and aborting the request. Returns HTTP error 400 and the message "Malformed XML, tag not closed."


Adam-:
Fixed in 3cb9e0b97c15e2d997d519b4bfc1821252c2384d

PR #195, opened 2017-07-24
Allow nickserv/update to be called from XMLRPC

kaecyra:
This PR allows the update command to lookup the online user from the supplied username context in an XMLRPC call, instead of silently failing.

### DIFF
--- a/modules/commands/ns_update.cpp
+++ b/modules/commands/ns_update.cpp
@@ -23,14 +23,14 @@ class CommandNSUpdate : public Command
 	{
 		User *u = source.GetUser();
 
-        // XMLRPC: Try to look up the user from their nick
-        if (!u)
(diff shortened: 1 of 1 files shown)


Adam-:
Maybe this type of functionality (where it executes as the user if they are online) should be built into m_xmlrpc_main instead?

kaecyra:
Thats where I started, but I was worried you'd view it as too big of a change. If you're up for that, its definitely the saner option imo. Shall I do that instead?

Adam-:
I'm not sure how big it would be. Try it and see?

kaecyra:
The code itself is miniscule. The effect would be large :) 
I already made the change, will do some testing and see how it goes. New PR when I have something to show.

PR #196, opened 2017-07-25
Allow calls to XMLRPC to impersonate supplied user, if online

kaecyra:
Some calls require a user object to work properly (NS/UPDATE, NS/GROUP, etc). This PR allows m_xmlrpc_main to lookup user objects by nickname and impersonate that user if they are online, allowing these calls to work.

### DIFF
--- a/modules/m_xmlrpc_main.cpp
+++ b/modules/m_xmlrpc_main.cpp
@@ -108,16 +108,7 @@ class MyXMLRPCEvent : public XMLRPCEvent
 				}
 				reply(out);
 
-				User *u = NULL;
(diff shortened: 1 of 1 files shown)


Adam-:
Whats with all this verbose logging? I don't see why it would need any more logging than normal

kaecyra:
Since its new functionality, I felt it might be useful if others were using it to see what it was doing. I don't have a problem removing it if you disapprove.

Adam-:
This should just be

```
User *u = User::Find(user, true);
CommandSource source(user, u, na ? *na->nc : NULL, &reply, bi);
```

PR #197, opened 2017-07-26
2.0: Fix sending incorrect RLines to InspIRCd

genius3000:
Currently a Regex AKILL is sent with a malformed mask to InspIRCd as an RLine.
The mask is sent as `/n!u@h#real name/`

InspIRCd expects a mask of `n!u@h\sreal\sname`, so we need to remove the enclosing slashes (/.../), change the '#' separator to '\s', and change any literal spaces to '\s' and then it creates a proper RLine.

This fix only applies to the modified mask sent to InspIRCd when RLine is available and assumes that no spaces will exist in the mask unless a real name is provided (n!u@h should never have spaces).
I have tested this and it does set the correct RLines now.


modules/protocol/inspircd12.cpp
```diff
@@ -105,13 +105,26 @@ class InspIRCd12Proto : public IRCDProto
...
-			size_t h = x->mask.find('#');
+			if (mask[0] == '/' && mask[mask.length() - 1] == '/')
+				mask = mask.substr(1, x->mask.length() - 2);
+			size_t h = mask.find('#');
 			if (h != Anope::string::npos)
-				mask = mask.replace(h, 1, ' ');
+			{
+				mask = mask.replace(h, 1, "\\s");
+				size_t s = mask.find(' ');
+				while (s != Anope::string::npos)
```
Adam-: There is a replace_all_cs, does that not do this?

genius3000: Yeah totally forgot about that, it for sure does this; replaced & updated.

PR #198, opened 2017-09-25
Replace file_get_contents with curl

NikosPapakonstantinou:
For security reasons & faster responses.

### DIFF
--- a/docs/XMLRPC/xmlrpc.php
+++ b/docs/XMLRPC/xmlrpc.php
@@ -40,20 +40,19 @@ class AnopeXMLRPC
      */
     public function run($name, $params)
     {
-        $xmlquery = xmlrpc_encode_request($name, $params);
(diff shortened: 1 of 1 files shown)


docs/XMLRPC/xmlrpc.php
```diff
@@ -41,18 +41,17 @@ public function __construct($host)
...
     {
         $xmlquery = xmlrpc_encode_request($name, $params);
-        $context = stream_context_create(["http" => [
-            "method" => "POST",
-            "header" => "Content-Type: text/xml",
-            "content" => $xmlquery]]);
-
-        $inbuf = file_get_contents($this->host, false, $context);
+        $ch = curl_init();
+        curl_setopt($ch, CURLOPT_URL, $host);
```
key2peace: Shouldn't that be:
curl_setopt($ch, CURLOPT_URL, $this->host);

PR #202, opened 2017-11-29
Add InspIRCd 1.2 and 2.0 SVSJOIN workaround

CuleX:
`SVSJOIN` is broken on InspIRCd 1.2 and 2.0 for keyed channels (see https://github.com/inspircd/inspircd/pull/892). This is a workaround by issuing an `INVITE` first if necessary. It's still *slightly* confusing to users, but at least they'll actually get joined to the channel.

This commit also makes sure the key parameter is always passed properly to `SendSVSJoin()`.

### DIFF
--- a/modules/commands/ns_recover.cpp
+++ b/modules/commands/ns_recover.cpp
@@ -246,13 +246,16 @@ class NSRecover : public Module
 				{
 					Channel *c = Channel::Find(it->first);
 					const Anope::string &cname = it->first;
+					Anope::string key;
 					++it;
 
 					/* User might already be on the channel */
 					if (u->FindChannel(c))
 						this->OnJoinChannel(u, c);
-					else if (IRCD->CanSVSJoin)
-						IRCD->SendSVSJoin(NickServ, u, cname, "");
+					else if (IRCD->CanSVSJoin) {
+						c->GetParam("KEY", key);
+						IRCD->SendSVSJoin(NickServ, u, cname, key);
+					}
 				}
 		}
 
--- a/modules/commands/os_svs.cpp
+++ b/modules/commands/os_svs.cpp
@@ -89,6 +89,7 @@ class CommandOSSVSJoin : public Command
 
 		User *target = User::Find(params[0], true);
 		Channel *c = Channel::Find(params[1]);
+		Anope::string key;
 		if (target == NULL)
 			source.Reply(NICK_X_NOT_IN_USE, params[0].c_str());
 		else if (source.GetUser() != target && (target->IsProtected() || target->server == Me))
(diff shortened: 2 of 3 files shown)


Robby-:
It has to be noted that in order for this `INVITE` workaround to even work, `<options:invitebypassmodes="yes">` must be set in inspircd.conf, which, granted, is the default, but still. In light of this, I'm not so sure if we should add this part of the patch to Anope, because it will do nothing useful for networks where that setting is set to `no`, and will probably only add to the confusion.

CuleX:
Anope already requires a number of configuration items (see the `CAPAB` handling requiring modules to be loaded on InspIRCd). Unfortunately, this particular setting cannot be detected from `CAPAB` output.

---

**With `INVITE` workaround**: There are two possibilities:

1. options:invitebypassmodes is yes: `SVSJOIN` works and thus the user-facing NickServ `AJOIN` and `RECOVER` commands work. This is also the default behavior.
2. options:invitebypassmodes is no: There will be two messages that make no sense to the user. If pasted to an IRC operator, they may be able to trace it back to this option and maybe change the IRCd configuration files.

**Without INVITE workaround**: There will be one message that make no sense to the user. If pasted to an IRC operator, they may be able to trace it back to this issue. There's no way to actually fix this.

---

The former option has one way to fix the issue for the affected user(s); the latter doesn't. This makes me consider it'd be more worthwhile than not merging. I would be fine with writing this detail into the `INSTALL` or something, too.

KoraggKnightWolf:
On InspIRCd 2.0.26 with Anope 2.0.6 which has this pull request implemented, whenever anyone calls the recover command, also through /ns ghost or /ns release there is the chance of services quitting with Connection closed. After asking on irc.anope.org #anope it was brought to my attention that 'c' could perhaps be NULL which would cause a crash (observed netsplits). The chance of it occuring would also explain why not every instance of its utlization crashed it.
I managed to circumvent the issue by reversing the steps of this pull request and we have not experienced any more such crashes since.
This would appear to be a rather drastic bug if any usage of /ns recover (and /ns ghost ns release if enabled) could cause services to crash whether deliberately with such intentino or not.
I hope that this bug report can help to track things down, the possible reason was explained to me by @genius3000 
Thank you for your time and effort in advance.

Regards,
Koragg

Lord255:
"As mentioned previously InspIRCd v2 is now deprecated and will no longer receive security updates after 2020-05-31 2020-12-31."

is this valid for v3 as well?

SadieCat:
I don't think this is valid for InspIRCd v3 and, either way, it doesn't edit the v3 protocol module.

PR #204, opened 2017-12-18
Memo deletion: add new privilege and some logging

LeCoyote:
Any user with the MEMO privilege can delete a channel memo without leaving a trace. These two commits add a MEMODEL privilege and generates logs every time a channel memo is deleted.

### DIFF
--- a/data/chanserv.example.conf
+++ b/data/chanserv.example.conf
@@ -519,7 +519,7 @@ privilege
 /*
  * MEMO privilege.
  *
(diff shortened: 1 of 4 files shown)


Adam-:
I don't want to see this level of behavior change in 2.0. I also think a privilege specifically for channel memo deletion is silly. Instead I would just make memo deletion log, as it was probably an oversight in the first place.

LeCoyote:
Ok, I'll do a new PR with just the logging.

PR #207, opened 2018-01-07
Add operserv/raw command

bonnedav:
I think you should re-add the raw command you had it in 1.x an it is helpful for testing the ircd. I did not make this myself I don't know who did, I just found it on the internet and fixed the grammar/spelling and made it so you need to be super admin to use it.

### DIFF
--- a/modules/commands/os_raw.cpp
+++ b/modules/commands/os_raw.cpp
@@ -1,0 +1,46 @@
+#include "module.h"
+#include <numeric>
+
+class CommandOsRaw : public Command
+{
+ public:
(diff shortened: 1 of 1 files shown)


SaberUK:
The raw command was removed because of idiots using it, breaking their networks, and then blaming Anope for being buggy. I don't think it should be brought back. People who *really* need this kind of functionality know how to add it.

DukePyrolator:
We don't want any os_raw functionality in anope.
We also don't accept it as a third party module on modules.anope.org. So don't try it. ;-)
Pull request closed. Have a nice day.

PR #208, opened 2018-01-22
Update anope.example.conf

TheMythPT:
Updated the configuration example to match the new syntax of listen and link blocks of UnrealIRCd 4.x, since 3.2 is deprecated and no longer supported by UnrealIRCd team.

### DIFF
--- a/data/anope.example.conf
+++ b/data/anope.example.conf
@@ -134,22 +134,31 @@ define
  *     <uline server="services.localhost.net" silent="yes">
  *     <bind address="127.0.0.1" port="7000" type="servers">
  *
- * An example configuration for UnrealIRCd that is compatible with the below uplink
+ * An example configuration for UnrealIRCd 4.x (3.2 series are deprecated) that is compatible with the below uplink
  * and serverinfo configuration would look like:
+ *
+ *	listen {
+ *	     ip 127.0.0.1;
+ *	     port 7000;
+ *	     options {
+ *		serversonly;
+ *	     };
+ *	};
  *
  *     link services.localhost.net
  *     {
- *          username *;
- *          hostname *;
- *          bind-ip "127.0.0.1";
- *          port 7000;
- *          hub *;
- *          password-connect "mypassword";
- *          password-receive "mypassword";
+ *          incoming {
+ *          	mask *@127.0.0.1;
+ *          };
+ *         outgoing { 
+ *          	bind-ip 127.0.0.1;
+ *          	hostname 127.0.0.1;
+ *          	port 7000;
+ *          };
+ *          password "mypassword";
  *          class servers;
- *     };
+ *      };
  *     ulines { services.localhost.net; };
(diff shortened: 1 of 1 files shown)


data/anope.example.conf
```diff
@@ -134,22 +134,31 @@ define
...
+ *          };
+ *         outgoing { 
+ *          	bind-ip 127.0.0.1;
+ *          	hostname 127.0.0.1;
+ *          	port 7000;
+ *          };
+ *          password "mypassword";
  *          class servers;
- *     };
+ *      };
```
SaberUK: There is extraneous whitespace here.

data/anope.example.conf
```diff
@@ -134,22 +134,31 @@ define
  *     <uline server="services.localhost.net" silent="yes">
  *     <bind address="127.0.0.1" port="7000" type="servers">
  *
- * An example configuration for UnrealIRCd that is compatible with the below uplink
+ * An example configuration for UnrealIRCd 4.x (3.2 series are deprecated) that is compatible with the below uplink
```
SaberUK: You probably don't need to mention that 3.2 is deprecated.

data/anope.example.conf
```diff
@@ -134,22 +134,31 @@ define
...
+ *     serversonly;
+ *    };
+ * };
+ *
+ * link services.localhost.net
+ * {
+ *   incoming {
+ *    mask *@127.0.0.1;
+ *   };
+ *   outgoing {
```
Adam-: Is outgoing necessary in the link block? You can't /connect services

data/anope.example.conf
```diff
@@ -134,22 +134,31 @@ define
...
+ * listen {
+ *   ip 127.0.0.1;
+ *   port 7000;
+ *    options {
+ *     serversonly;
+ *    };
+ * };
+ *
+ * link services.localhost.net
+ * {
```
Adam-: This should probably be on the same line as "link" for consistency

data/anope.example.conf
TheMythPT: And you're correct. For services the outgoing block isn't needed. I also went with `*@127.0.0.1` in the incoming because I add a few issues with connection denied because it said that anope@127.0.0.1 wasn't a valid link block or something like that.

data/anope.example.conf
TheMythPT: The listen block is fine. I have it like that and as per UnrealIRCd docs this is how it should be.

TheMythPT:
I'll fix it in a bit. Just returning home.

data/anope.example.conf
TheMythPT: Oh sorry, I misunderstood the point here.
Perhaps the `{` should be on the same line, yes. but I just copied the example from UnrealIRCd docs.

Adam-:
Added in 67b7fed9e

PR #209, opened 2018-01-28
modules/dns: NotifySocket takes a bool parameter, not int address family

nomis:
gcc warning: ?: using integer constants in boolean context, the expression will always evaluate to ‘true’ [-Wint-in-bool-context]


Adam-:
Thanks. I have applied this to 2.0 too.

PR #222, opened 2018-06-12
Unreal4 does not allow SWHOIS to come from a user source.

bonnedav:


Adam-:
Alternatively, don't set the source to be a user? I do not think Anope uses swhois functionality anywhere currently and someone may want to specify it being from a specific server.

bonnedav:
I did it like this so it can still come from a user source on other ircds. I guess I don't see why someone would want it to be from a different server. Besides, how else should we make it work for Unreal if it is called with a user source because some other ircds do support it. Many this function can be changed to see if the source is a user or a server and use the source only if it is a server? Would there be any point in that? Is it done anywhere else?

PR #225, opened 2018-07-19
Handle S2S AWAY messages with a timestamp

genius3000:
InspIRCd 2.0 sends a timestamp in its S2S AWAY message since
https://github.com/inspircd/inspircd/commit/b1002b7df55f1aec5bfce48ecd07d6d326174c95
Syntax of `AWAY <timestamp> :<away msg>`
Catch this message and remove the timestamp parameter before
sending it to the core AWAY handler.

**Tested** successfully.

**Edit**: Changed to a modification of the `inspircd20` protocol instead of core.

### DIFF
--- a/modules/protocol/inspircd20.cpp
+++ b/modules/protocol/inspircd20.cpp
@@ -369,6 +369,20 @@ class ChannelModeRedirect : public ChannelModeParam
 	}
 };
 
+struct IRCDMessageAway : Message::Away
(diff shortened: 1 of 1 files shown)


genius3000:
Looked at that, but it would be duplicating the code for one tiny difference. Granted it would be safer incase any others add parameters after the awaymsg.

PR #226, opened 2018-09-23
Fixed some misspellings in Italian Translation

Dragone2:
I have fixed some misspellings in Italian translation file.


Adam-:
Thanks!

PR #227, opened 2018-10-11
Polish translation reworked and completed.

k4bek4be:
This was greatly incomplete since the introduction of anope 2.0.


Adam-:
Thanks!

PR #229, opened 2018-10-28
Update chanstats.example.conf

TheMythPT:
Missing permissions that otherwise allow any registered user to turn on/off chanstats option to any other user


Robby-:
Merged via 9069bf4e2f28cae7d12cd0f90bcb2b1d99803c71.
Thank you!

PR #234, opened 2019-01-25
Allow customising the enabled SASL mechanisms.

SadieCat:


modules/m_sasl.cpp
```diff
@@ -320,23 +320,43 @@ class ModuleSASL : public Module
...
-		sasl(this), plain(this), external(NULL)
+	template<typename Mech>
+	void ToggleMech(Mech *mech, bool enable)
 	{
-		try
+		if (enable)
+		{
+			try
+			{
+				mech = new Mech(this);
```
Adam-: Are you sure about this? I don't think it would change the pointer in the module unless `Mech` is a reference?

modules/m_sasl.cpp
```diff
@@ -320,23 +320,43 @@ class ModuleSASL : public Module
...
+			}
+			catch (ModuleException &) { }
+		}
+		else if (!enable && mech)
 		{
-			external = new External(this);
-			CheckMechs();
+			delete mech;
```
Adam-: this will double free

PR #235, opened 2019-02-09
Allow post-handshake SASL in UnrealIRCd 4.2.2 and up.

syzop:
This updates the unreal4 protocol module to work with both pseudo-id's
and real UID's. Something that will also be necessary once UnrealIRCd
gets rid of pseudo-id's altogether.

### DIFF
--- a/modules/protocol/unreal4.cpp
+++ b/modules/protocol/unreal4.cpp
@@ -381,7 +381,9 @@ class UnrealIRCdProto : public IRCDProto
 			if (!s)
 				return;
 			distmask = s->GetName();
-		} else {
+		}
+		else
+		{
 			distmask = message.target.substr(0, p);
 		}
 
--- a/modules/protocol/unreal4.cpp
+++ b/modules/protocol/unreal4.cpp
@@ -399,7 +399,9 @@ class UnrealIRCdProto : public IRCDProto
 			if (!s)
 				return;
 			distmask = s->GetName();
-		} else {
+		}
+		else
+		{
 			distmask = uid.substr(0, p);
 		}
 		UplinkSocket::Message(Me) << "SVSLOGIN " << distmask << " " << uid << " " << acc;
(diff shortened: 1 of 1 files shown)


syzop:
Until now anope assumed pseudo-id's with the ! separator etc (simply copied from 3.2.x, which is understandable as we did so as well).
This PR changes it so that if there is no "!" character in the target then we do a server hunt for the first 3 characters. This is similar to what the charybdis protocol module does, but then in a backward compatible way.

If you need additional information or have doubts about part of the code, let me know. I have of course tested it with both pre-auth (with PUID) and post-auth (with UID), but I must also say I am not very familiar with the anope source code nor with C++ in general.

Thanks!

modules/protocol/unreal4.cpp
```diff
@@ -373,18 +373,36 @@ class UnrealIRCdProto : public IRCDProto
...
+		Anope::string distmask;
+
 		if (p == Anope::string::npos)
-			return;
+		{
+			Server *s = Server::Find(message.target.substr(0, 3));
+			if (!s)
+				return;
+			distmask = s->GetName();
+		} else {
```
Adam-:
```suggestion
		}
		else
		{
```

modules/protocol/unreal4.cpp
```diff
@@ -373,18 +373,36 @@ class UnrealIRCdProto : public IRCDProto
...
+
 		if (p == Anope::string::npos)
-			return;
-		UplinkSocket::Message(Me) << "SVSLOGIN " << uid.substr(0, p) << " " << uid << " " << acc;
+		{
+			Server *s = Server::Find(uid.substr(0, 3));
+			if (!s)
+				return;
+			distmask = s->GetName();
+		} else {
```
Adam-:
```suggestion
		}
		else
		{
```

jesopo:
:shipit:

PR #236, opened 2019-02-09
Update nickserv.example.conf regarding "addaccessonreg" feature

PeGaSuS-Coder:
When set to `yes` this breaks the `killprotect` functionality, because a user is expected to turn into guest after the specified amount of time.
Also new users that register their nicknames will wonder why they can't join a +R channel but they didn't change their nick to Guest*

### DIFF
--- a/data/nickserv.example.conf
+++ b/data/nickserv.example.conf
@@ -261,7 +261,7 @@ module
 	 * being recognized, unless they manually add an address to the access list of their account.
 	 * This directive is optional.
 	 */
-	addaccessonreg = yes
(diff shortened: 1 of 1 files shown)


KoraggKnightWolf:
Not to mention it is a possible security consideration as it always uses the ident and the non unique part of a hostmask so anyone with the same ISP (or e.g. VPS service etc) and the same ident (often configurable per client) could fake being that user (easy to deceive newbies who do not know to confirm via /WHOIS looking for "is logged in as").

Regards,

Koragg

PR #237, opened 2019-02-13
Update anope.en_US.po

TheMythPT:
Add a better explanation about successor and ownership changes


Robby-:
Fixed in e567cfda360d7f374698e5e7f5d12cc3a6db357c. Language files will be regenerated later to accommodate this change.

PR #238, opened 2019-02-13
Update cs_set.cpp

TheMythPT:
Add a better explanation about successor and ownership changes


Robby-:
Fixed in e567cfda360d7f374698e5e7f5d12cc3a6db357c.

PR #240, opened 2019-03-30
Mark pseudoclients as bots on InspIRCd if the bot mode exists.

SaberUK:
This will allow their messages to be marked as originating from bots on InspIRCd v3 (ref: https://github.com/inspircd/inspircd/commit/180b8b6ab194fd74cb5c64fc3ec0775eb542451a).


Adam-:
If services disconnect/reconnect this might cause the mode to be added multiple times?

PR #242, opened 2019-06-05
Add support for the InspIRCd 1205 protocol.

SaberUK:
This should work fine but it'll require a bit more testing.

Note: this depends on #241.


SaberUK:
This should be ready to merge once #214 is merged.

Adam-:
This also requires an update to `docs/README` to include the new supported inspircd version

PR #247, opened 2019-11-12
inspircd3: Add the SSL ExtensibleItem for Users.

genius3000:
This ext item is how SSL status is known for a user on InspIRCd. This worked with InspIRCd 3 using the `inspircd20` protocol because it inherited from `inspircd12`, which declared this item. The relevant code for `Extend` already exists.

Fixes bug 1728.


SaberUK:
Nice catch!

PR #253, opened 2020-01-05
Small fix to support translation in m_chanstats + Update to Italian translation

Dragone2:
Hi,
I've added a small fix to m_chanstats.cpp to support translations and I've updated the Italian translation with some fixes to the strings of chanstats.


Robby-:
Merged as 83eea993171642b522d579ec516e53051dfcef86 and d319fc6eb547acd47c74db6a706ceb525d371d49.
Thank you for for your contribution.

PR #254, opened 2020-03-26
Fix the handling of topic messages with InspIRCd v3.

genius3000:
The "normal" `FTOPIC` message is only 4 parameters and uses the source as the setter. During burst or in reply to `RESYNC` 5 parameters are supplied, with the setter being supplied before the topic.
Reference: https://github.com/inspircd/inspircd/blob/insp3/src/modules/m_spanningtree/ftopic.cpp#L74-L93

Credit to Dimitris for spotting this bug and reporting it in #anope (no idea if they are on GitHub).


SadieCat:

PR #256, opened 2020-04-17
Update unreal4.cpp

PeGaSuS-Coder:
While using `/HS OFF`, if the IRC client has no support for CHGHOST capability, the user IP is exposed for a split of a second.
This PR aims to fix that.

### Using /HS OFF #
* Test (~vitor@PTirc/Users/Test) has left (Changing host)
* Test (~vitor@xx.xxx.xx.xx) has joined
* Test (~vitor@xx.xxx.xx.xx) has left (Changing host)
* Test (~vitor@DAB1F162.F8B482F.58EF52D7.IP) has joined

### Using /HS OFF #
* Test (~vitor@PTirc/Users/Test) has left (Changing host)
* Test (~vitor@DAB1F162.F8B482F.58EF52D7.IP) has joined

Some input from @syzop would be appreciated.

Cheers

### DIFF
--- a/modules/protocol/unreal4.cpp
+++ b/modules/protocol/unreal4.cpp
@@ -78,18 +78,7 @@ class UnrealIRCdProto : public IRCDProto
 	void SendVhostDel(User *u) anope_override
 	{
 		BotInfo *HostServ = Config->GetClient("HostServ");
-		/* Should we unset the cloak (umode +x)? If the IRC client has no support to the CHGHOST cap
-		 * then the user real IP is displayed for a fraction of a second which is bad
-		 * For security, we only remove the vhost (umode -t) without touching the cloak
-		 * Line is only commented as it might be useful in a near future
-		 */
-		// u->RemoveMode(HostServ, "CLOAK");
 		u->RemoveMode(HostServ, "VHOST");
-		/* Because the change above this is no longer needed
-		 * Commented as it might be useful in a near future
-		 */
-		// ModeManager::ProcessModes();
-		// u->SetMode(HostServ, "CLOAK");
 	}
 
 	void SendAkill(User *u, XLine *x) anope_override
@@ -215,11 +204,6 @@ class UnrealIRCdProto : public IRCDProto
 			UplinkSocket::Message(Me) << "CHGHOST " << u->GetUID() << " " << vhost;
 		// Internally unreal sets +xt on chghost
 		BotInfo *bi = Config->GetClient("HostServ");
-		/* If user is already uncloaked (umode -x) due to personal choice or IRCd config
-		 * we shouldn't force the cloak in that case. Line commented because it might be useful in a near future
-		 */
-		// u->SetMode(bi, "CLOAK");
-		/* This is where se set the vHost */
 		u->SetMode(bi, "VHOST");
 	}
 
--- a/modules/protocol/unreal4.cpp
+++ b/modules/protocol/unreal4.cpp
@@ -78,7 +78,18 @@ class UnrealIRCdProto : public IRCDProto
 	void SendVhostDel(User *u) anope_override
 	{
 		BotInfo *HostServ = Config->GetClient("HostServ");
+		/* Should we unset the cloak (umode +x)? If the IRC client has no support to the CHGHOST cap
+		 * then the user real IP is displayed for a fraction of a second which is bad
+		 * For security, we only remove the vhost (umode -t) without touching the cloak
+		 * Line is only commented as it might be useful in a near future
+		 */
+		// u->RemoveMode(HostServ, "CLOAK");
 		u->RemoveMode(HostServ, "VHOST");
+		/* Because the change above this is no longer needed
+		 * Commented as it might be useful in a near future
+		 */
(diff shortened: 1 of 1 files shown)


PeGaSuS-Coder:
Worth to mention that this only happens if **`set::allow-userhost-change`** is set to **force-rejoin**, sine the default `**`always`** can lead to client desyncs and if the IRC client has no support to CHGHOST cap

Adam-:
Does this work? This used to not work, at least on Unreal 3... you could probably just remove this `u->SetMode(HostServ, "CLOAK");` call then too?

PeGaSuS-Coder:
I've tested this in my network and it works.

I rather keep `u->SetMode(HostServ, "CLOAK");`, just in case?

PeGaSuS-Coder:
Also be aware that this was a quick fix without any knowledge of programming nor programming skills. I was even surprised that this worked. Cheers

Adam-:
Otherwise, if you were -x and did /hs on then /hs off it would cloak you, which seems odd?

PeGaSuS-Coder:
I believe that all the bits mentioned above are covered with this rebase.
Although I'd like to hear something from @syzop first, this looks good to me.
Cheers.

PeGaSuS-Coder:
I've pushed the changes into my Anope instance and all seems to work as expected.

SadieCat:
You don't need to leave old code commented out. If it needs to be restored it can be pulled out of Git history.

PeGaSuS-Coder:
Do you prefer for me to wipe them then?

Adam-:
You should remove old code instead of commenting it out, this is why git and commit messages exist.

SadieCat:
@syzop Is this accurate behaviour for UnrealIRCd 4+?

syzop:
Yes, you can send ```-t``` instead of ```-xt``` to unset the vhost and keep the user cloaked.
We already supported that for a couple of years in ```MODE``` (so for user doing it onto themselves), and after fixing https://bugs.unrealircd.org/view.php?id=5111 it is supported in ```SVSMODE``` / ```SVS2MODE``` in UnrealIRCd 4.2.0 and above (https://github.com/unrealircd/unrealircd/commit/858aaa774a409b65ee0e3ddefddbb7e63b86bf1d). Version 4.2.0 was released on September 30, 2018. It's in UnrealIRCd 5.x as well (since 5.0.0).

The bottom half of the patch does not seem correct though. It says ```// Internally unreal sets +xt on chghost``` and that is correct. So the line ```u->SetMode(bi, "CLOAK");``` should not be removed, it should be kept. In other words, the ```SendVhost``` function should be left unchanged.

So yeah, with that last change, it sounds good to me. But... i did not test this patch, and I think it should be tested first (@PeGaSuS-Coder  ?) so mostly commenting on the -t behavior.

PeGaSuS-Coder:
Okay. I've tested this and I'm kinda confused with this behaviour:

#### With umode +x
User joins the channel, with a vHost already set:
```
[01/11/2020 - 13:59:14] Test (~Wockets@PTirc/Users/Test) has joined
[01/11/2020 - 13:59:14] irc.ptirc.org sets modes [#spamtest +o Test]
```
User do /hs off:
```
[01/11/2020 - 14:03:06] Test (~Wockets@PTirc/Users/Test) has left (Changing host)
[01/11/2020 - 14:03:06] Test (~Wockets@DAB1F162.F8B482F.58EF52D7.IP) has joined
[01/11/2020 - 14:03:06] irc.ptirc.org sets modes [#spamtest +o Test]
```
User do /hs on:
```
[01/11/2020 - 14:04:45] Test (~Wockets@DAB1F162.F8B482F.58EF52D7.IP) has left (Changing host)
[01/11/2020 - 14:04:45] Test (~Wockets@PTirc/Users/Test) has joined
[01/11/2020 - 14:04:45] irc.ptirc.org sets modes [#spamtest +o Test]
```
So far all seems to work properly when user has `umode +x` set from the beginning

#### Unset umode +x
User joins the channel, with a vHost already set:
```
[01/11/2020 - 14:08:49] Test (~Wockets@PTirc/Users/Test) has joined
[01/11/2020 - 14:08:49] ChanServ sets modes [#spamtest +o Test]
```
User do /mode nick -x:
```
[01/11/2020 - 14:10:32] Test (~Wockets@PTirc/Users/Test) has left (Changing host)
[01/11/2020 - 14:10:32] Test (~Wockets@41.53.152.77.rev.sfr.net) has joined
[01/11/2020 - 14:10:32] irc.ptirc.org sets modes [#spamtest +o Test]
```
This is a weird behaviour. Why is the IRCd removing the vHost when we only asked to remove the cloak?
After this, using /hs on or /hs off behaves like if the user had `umode +x`.

Since I'm unable to make it work as I would like to, I'll close this PR in the next days unless someone can tell me what to edit.

Cheers

PeGaSuS-Coder:
After a few more tests, some erratic behaviour seems to have been introduced by my PR. Closing.

syzop:
Sorry but this PR is too much noise, it's so bad. I suggest you close it indeed. I will deal with it myself later but will probably wait until 4.x is EOL'd.

PR #257, opened 2020-05-14
Rename current unreal protocol module to "unreal"

k4bek4be:
This will probably help users to avoid mistakenly using the legacy module.


SadieCat:
This breaks all existing config files for upgrading UnrealIRCd users.

A better fix might be to add "(for 4.0 or newer)" here: https://github.com/anope/anope/blob/2.0/data/example.conf#L264

SadieCat:
This breaks all existing config files for upgrading users.

A better fix might be to add "(for 4.x or newer)" here: https://github.com/anope/anope/blob/2.0/data/example.conf#L264

k4bek4be:
This makes sense, thanks. I'll submit another PR.

PR #259, opened 2020-06-09
Update ns_drop.cpp

PeGaSuS-Coder:
Automatically remove the vHost of a user as soon as they drop the nickname.

I don't know how accurate this changes are, but seems to work on my network (UnrealIRCd 5).

Cheers.

### DIFF
--- a/modules/commands/ns_drop.cpp
+++ b/modules/commands/ns_drop.cpp
@@ -45,10 +45,14 @@ class CommandNSDrop : public Command
 			source.Reply(_("You may not drop other Services Operators' nicknames."));
 		else
 		{
+			User *u = source.GetUser();
 			FOREACH_MOD(OnNickDrop, (source, na));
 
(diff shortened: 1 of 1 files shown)


modules/commands/ns_drop.cpp
```diff
@@ -45,10 +45,14 @@ class CommandNSDrop : public Command
 			source.Reply(_("You may not drop other Services Operators' nicknames."));
 		else
 		{
+			User *u = source.GetUser();
```
genius3000, line 48: `source` could be a services operator force dropping a nick. For future reference, `source.GetUser()` could also return NULL when a command is called programmatically (i.e., web panel (drop could be implemented someday) or a custom module, etc.)

PeGaSuS-Coder:
Honestly, this was made in a `guessing` basis, because I'm mostly unaware how all this works. 🤐

If you could tweak it to work properly, it would be great.

What I need is:
1. If the source is the nick and is logged in and the vHost is per account, logout the nick and remove the vHost only for that nick
2. If the source is an IRCop and the nick is logged in and vHost is per account, logout the nick and remove the vHost only for that nick

Thanks in advance!

PeGaSuS-Coder:
Closing this PR as it isn't working properly after a few more tests and I can't do better than this.

If anyone wants to take it from here, feel free to do so.

Cheers!

PR #263, opened 2020-09-09
Modernize ircd-hybrid protocol module. Add support for upcoming ircd-…

miwob:
…hybrid versions.

### DIFF
--- a/modules/protocol/hybrid.cpp
+++ b/modules/protocol/hybrid.cpp
@@ -12,6 +12,7 @@
 #include "module.h"
 
 static Anope::string UplinkSID;
+static bool UseSVSAccount = false;  // Temporary backwards compatibility hack until old proto is deprecated
 
 class HybridProto : public IRCDProto
 {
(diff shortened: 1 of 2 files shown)


modules/protocol/hybrid.cpp
```diff
@@ -408,32 +378,35 @@ struct IRCDMessagePong : IRCDMessage
...
 	{
 		/* Servers other than our immediate uplink are introduced via SID */
 		if (params[1] != "1")
 			return;
 
-		new Server(source.GetServer() == NULL ? Me : source.GetServer(), params[0], 1, params[2], UplinkSID);
+		if (params.size() == 5)
+			UplinkSID = params[2];
+
+		new Server(source.GetServer() == NULL ? Me : source.GetServer(), params[0], 1, params[params.size() - 1], UplinkSID);
```
SadieCat, line 394:
```suggestion
		new Server(source.GetServer() == NULL ? Me : source.GetServer(), params[0], 1, params.back(), UplinkSID);
```

miwob:
ircd-hybrid does RHOST since early 2017. Minimum supported version would be 8.2.23 I think.
What sourceless commands are you talking about specifically? Servers can add/remove *LINEs since about half a decade now or something.

Adam-:
I don't mind bumping the min version but currently the docs and the module claim to support 8.2.x and I think this would break that. Specifically I am talking about all of the ones in this diff.

miwob:
I've bumped the min version mentioned in the module. The docs don't seem to mention a specific version at all from what I have seen.

Adam-:
They do here https://github.com/anope/anope/blob/2.0/docs/README#L171

PR #267, opened 2020-10-23
Update nickserv.example.conf

PeGaSuS-Coder:
Disable the nickserv/set/display and nickserv/saset/display, which can be used to bypass extbans (in IRCds that support those) based on account name.

Cheers


SadieCat:
There are plenty of easier ways to ban evade than by this. If you want your IRCd to be able to ban by something immutable encourage the developers of it to add support for banning by uniqueid instead of display name.

If you still want this then you can always apply it to your own server.

PR #268, opened 2020-12-30
inspircd3: Enable support for channel SQLines

Techman:
Enables support for sending CBANs when a channel SQLine is made. Requires InspIRCd 3.8+ with m_cban configured to allow glob masks.

If m_cban is not configured with glob support, Anope will not send a CBAN and will fall back to its current system of enforcement.

### DIFF
--- a/modules/protocol/inspircd3.cpp
+++ b/modules/protocol/inspircd3.cpp
@@ -62,6 +62,7 @@ class InspIRCd3Proto : public IRCDProto
 		CanSetVHost = true;
 		CanSetVIdent = true;
 		CanSQLine = true;
+		CanSQLineChannel = true;
 		CanSZLine = true;
 		CanSVSHold = true;
 		CanCertFP = true;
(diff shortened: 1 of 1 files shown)


modules/protocol/inspircd3.cpp
```diff
@@ -1105,6 +1114,9 @@ struct IRCDMessageCapab : Message::Capab
 				}
 				else if (module.equals_cs("m_topiclock.so"))
 					Servers::Capab.insert("TOPICLOCK");
+				else if (module.find("m_cban.so") == 0)
+					if (module.length() > 10 && module.substr(10) == "glob")
```
SadieCat, line 1118: This is probably fine-

```suggestion
				else if (module.equals_cs("m_cban.so=glob"))
```

Techman: Heh, this is what happens when you follow the existing logic a little too closely :)

PR #269, opened 2020-12-31
Update nickserv.example.conf

PeGaSuS-Coder:
Since `addaccessonreg` is set to **no**, IMHO, this changes also add more security to newer installations/registered users.

Cheers


Robby-:
Please don't sneak `ns_keep_modes` into this. Only `killprotect`.

Robby-:
Merged as 7e08e7cb6c2548043e180b87a31577ffa1db65ab.

PR #272, opened 2021-01-09
hybrid.cpp: adjust IsIdentValid test to comply with upcoming ircd-hyb…

miwob:
…rid where user names may only consist of [A-Za-z0-9._-] and may not begin with '.', '-', or '_'.

As a side effect this fixes an issue where currently it is possible to introduce bots whose user name starts with (non-alnum) characters considered invalid	for ircd-hybrid leading to a services shutdown due to a nick introduction/kill loop.


SadieCat:
Merged, apologies for the delay and thanks for your contribution! 👍

PR #279, opened 2021-11-14
data/global.example.conf: Correction

siniStar7boy:
The previous default example (which is in use on many Anope networks) refers to Services in 2nd person, then 1st person - in the same sentence.


SadieCat:

PR #282, opened 2021-12-20
Fix missing spaces in de_DE

henk84:


SadieCat:
#281 will get applied to 2.1 during the next branch merge. Thanks anyway though.

PR #286, opened 2022-01-08
Use utf-8 encoding for outgoing email

k4bek4be:


src/mail.cpp
```diff
@@ -41,9 +41,10 @@ void Mail::Message::Run()
 	else
 		fprintf(pipe, "To: \"%s\" <%s>\n", mail_to.c_str(), addr.c_str());
 	fprintf(pipe, "Subject: %s\n", subject.c_str());
+	fprintf(pipe, "Content-Type: text/plain; charset=UTF-8;\n");
+	fprintf(pipe, "Content-Transfer-Encoding: 8bit\n");
 	fprintf(pipe, "\n");
 	fprintf(pipe, "%s", message.c_str());
-	fprintf(pipe, "\n.\n");
```
SadieCat, line 46: This is required to terminate the message according to RFC821 and is relied on by `anopesmtp` on Windows. Can you restore it?

k4bek4be: With sendmailpath configured to use `ssmtp`, these final two newlines and a dot appear inside the body of message. Maybe it would be a good idea to make it configurable then?

SadieCat: Are you using `ssmtp` directly or are you using its `sendmail`-compatible wrapper? You should be using the latter.

k4bek4be: I think I have tested both `/usr/sbin/ssmtp -t` and `/usr/sbin/sendmail -t` (being just a symlink). For now we have Anope configured for `ssmtp`. We may have used only some ancient version of ssmtp though.

I'll replace the footer for now and investigate that later.

PR #287, opened 2022-01-08
Add a "Network Service" (RPL_WHOISOPERATOR) line to WHOIS replies

k4bek4be:


SadieCat:
Will this cause a double 313 numeric on IRCds which mark their pseudoclients as an operator?

k4bek4be:
I would expect IRCds not to add anything to a remote WHOIS output, that's what the remote query is for: to make another server generate all replies. I don't know peculiarities of all S2S protocols though.

PR #289, opened 2022-02-08
Update unreal4.cpp

PeGaSuS-Coder:
This fixes the bug reported on https://bugs.unrealircd.org/view.php?id=6068

As Jobe stated, Anope was sending -xt and then a separate +x, instead of simply sending a -t, this resulted in 2 separate host change events and thus 2 separate part+join resyncs.

Previous behavior on **/hs off**:
```
[08/02/2022 - 00:16:02] * LocalGhost (uid376365@PTirc/Users/James) has left (Changing host)
[08/02/2022 - 00:16:02] * LocalGhost (uid376365@id-376365.uxbridge.irccloud.com) has joined
[08/02/2022 - 00:16:02] * irc3.ptirc.org sets modes [#Portugal +o LocalGhost]
[08/02/2022 - 00:16:02] * LocalGhost (uid376365@id-376365.uxbridge.irccloud.com) has left (Changing host)
[08/02/2022 - 00:16:02] * LocalGhost (uid376365@32E4615F:15F25CF6:80AEFEA0:IP) has joined
[08/02/2022 - 00:16:02] * irc3.ptirc.org sets modes [#Portugal +o LocalGhost]
```

Behavior after the fix:
```
[08/02/2022 - 01:06:30] * LocalGhost (uid376365@PTirc/Users/James) has left (Changing host)
[08/02/2022 - 01:06:30] * LocalGhost (uid376365@32E4615F:15F25CF6:80AEFEA0:IP) has joined
[08/02/2022 - 01:06:30] * irc3.ptirc.org sets modes [#Portugal +o LocalGhost]
```

### DIFF
--- a/modules/protocol/unreal4.cpp
+++ b/modules/protocol/unreal4.cpp
@@ -82,7 +82,6 @@ class UnrealIRCdProto : public IRCDProto
 	void SendVhostDel(User *u) anope_override
 	{
 		BotInfo *HostServ = Config->GetClient("HostServ");
-		u->RemoveMode(HostServ, "CLOAK");
 		u->RemoveMode(HostServ, "VHOST");
 		ModeManager::ProcessModes();
 		u->SetMode(HostServ, "CLOAK");
--- a/modules/protocol/unreal4.cpp
+++ b/modules/protocol/unreal4.cpp
@@ -84,7 +84,6 @@ class UnrealIRCdProto : public IRCDProto
 		BotInfo *HostServ = Config->GetClient("HostServ");
 		u->RemoveMode(HostServ, "VHOST");
 		ModeManager::ProcessModes();
-		u->SetMode(HostServ, "CLOAK");
 	}
 
 	void SendAkill(User *u, XLine *x) anope_override


SadieCat:
@k4bek4be can you confirm this?

modules/protocol/unreal4.cpp
```diff
@@ -82,7 +82,6 @@ class UnrealIRCdProto : public IRCDProto
 	void SendVhostDel(User *u) anope_override
 	{
 		BotInfo *HostServ = Config->GetClient("HostServ");
-		u->RemoveMode(HostServ, "CLOAK");
 		u->RemoveMode(HostServ, "VHOST");
 		ModeManager::ProcessModes();
 		u->SetMode(HostServ, "CLOAK");
```
SadieCat, line 87: Do you need to re-set the cloak mode here if you're not removing it?

PeGaSuS-Coder:
**Tested on my network after Sadie's comment**  

I've turned [this](https://github.com/anope/anope/blob/2.0/modules/protocol/unreal4.cpp#L82-L89) bit into this:
```
void SendVhostDel(User *u) anope_override
	{
		BotInfo *HostServ = Config->GetClient("HostServ");
		u->RemoveMode(HostServ, "VHOST");
	}
```

After that, I've recompiled Anope and on IRC I've done `/OS modreload unreal4` which replied with:
```
[08/02/2022 - 08:36:42] -- OperServ (OperServ@services.ptirc.org): Module unreal4 reloaded.
```

Then I've connected with Hexchat and did `/cap req -chghost`


With nick LocalGhost I've done `/HS off` which showed this:
```
[08/02/2022 - 08:38:04] * LocalGhost (uid376365@PTirc/Staff/James) has left (Changing host)
[08/02/2022 - 08:38:04] * LocalGhost (uid376365@32E4615F:15F25CF6:80AEFEA0:IP) has joined
[08/02/2022 - 08:38:04] * irc2.ptirc.org sets modes [#Portugal +o LocalGhost]
```

If you need any other tests let me know.

Cheers

k4bek4be:
Unsetting +x (CLOAK) will unset +t (VHOST). Unsetting +t won't unset +x. Disabling vhost should not remove +x (as the PR author noticed). Re-setting +x after removing +t is just a policy thing, if we want to be extra careful with hiding user addresses. Normally, it will remain as it used to be before applying the vhost.

PeGaSuS-Coder:
So I've tested this in more detail.

Re-setting the cloak shouldn't be done, because on my testnet I can remove +x (with `/mode nick -x`).

**Examples with the current PR:**
 - /hs off
```
[08/02/2022 - 19:30:20] * James (WeeChat@LetsChat/Users/James) has left (Changing host)
[08/02/2022 - 19:30:20] * James (WeeChat@9F1CC1EA.E559752E.F7DD85C8.IP) has joined
```

- /mode James -x
```
[08/02/2022 - 19:30:34] * James (WeeChat@9F1CC1EA.E559752E.F7DD85C8.IP) has left (Changing host)
[08/02/2022 - 19:30:34] * James (WeeChat@ns3293111.ip-5-135-186.eu) has joined
```

- /hs on
```
[08/02/2022 - 19:30:43] * James (WeeChat@ns3293111.ip-5-135-186.eu) has left (Changing host)
[08/02/2022 - 19:30:43] * James (WeeChat@LetsChat/Users/James) has joined
```

- /hs off
    - Now I should rejoin  with the IP visible though, but I'm rejoining cloaked which is overriding my usermodes
```
[08/02/2022 - 19:30:48] * James (WeeChat@LetsChat/Users/James) has left (Changing host)
[08/02/2022 - 19:30:48] * James (WeeChat@9F1CC1EA.E559752E.F7DD85C8.IP) has joined
```

There's somewhere else where UnrealIRCd is re-setting the cloak (and I can't figure it out), because even if I make the changes to be like:
```
void SendVhostDel(User *u) anope_override
	{
		BotInfo *HostServ = Config->GetClient("HostServ");
		u->RemoveMode(HostServ, "VHOST");
	}
```
I'm still being re-cloaked and this should not happen since I've chosen to set myself -x.

k4bek4be:
I've investigated this more deeply. Looks like it's unrealircd's choice to always (re)set +x when **adding** the vhost, so it will stay set when the vhost is later unset.
https://github.com/unrealircd/unrealircd/blob/unreal60_dev/src/modules/chghost.c#L335
In my opinion, both lines `u->RemoveMode(HostServ, "CLOAK");` and `u->SetMode(HostServ, "CLOAK");` need removing to match current ircd behaviour.
Also it's apparently an unrealircd bug that neither the affected user nor the server which issued the CHGHOST is notified about that +xt mode change. Plase report it on the bugtracker.

modules/protocol/unreal4.cpp
PeGaSuS-Coder: This should be resolved now.

PR #293, opened 2022-04-28
Webcpanel: Add ability to CONFIRM accounts via NickServ tab

siniStar7:
This PR allows users who register, using the Anope Webcpanel to login and confirm their email address _(registration)_ with the emailed **CONFIRM code**.

### DIFF
--- a/modules/webcpanel/pages/nickserv/confirm.cpp
+++ b/modules/webcpanel/pages/nickserv/confirm.cpp
@@ -1,0 +1,26 @@
+/*
+ * (C) 2003-2022 Anope Team
+ * Contact us at team@anope.org
+ *
+ * Please read COPYING and README for further details.
+ */
+
+#include "../../webcpanel.h"
+
+WebCPanel::NickServ::Confirm::Confirm(const Anope::string &cat, const Anope::string &u) : WebPanelProtectedPage(cat, u)
+{
+}
+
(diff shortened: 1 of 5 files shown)


PeGaSuS-Coder:
I wonder if we could have this merged.

It is a great feature and very useful for a `sasl-only` network type, since you can't use SASL to identify to a non confirmed account.

Cheers!

PeGaSuS-Coder:
I've tried to use this PR and it worked fine.

@siniStar7, I'm wondering if it would be possible to add a way to check if the user is confirmed and if so, to hide the `Confirm`.

That's the only thing I think that can be improved. Other than that, great PR.

Cheers!

modules/webcpanel/pages/nickserv/confirm.cpp
```diff
@@ -0,0 +1,26 @@
...
+{
+}
+
+bool WebCPanel::NickServ::Confirm::OnRequest(HTTPProvider *server, const Anope::string &page_name, HTTPClient *client, HTTPMessage &message, HTTPReply &reply, NickAlias *na, TemplateFileServer::Replacements &replacements)
+{
+
+		std::vector<Anope::string> params;
+		params.push_back(message.post_data["code"]);
+
+		WebPanel::RunCommand(client, na->nc->display, na->nc, "NickServ", "nickserv/confirm", params, replacements);
```
SadieCat, line 20:
```suggestion
	std::vector<Anope::string> params;
	params.push_back(message.post_data["code"]);

	WebPanel::RunCommand(client, na->nc->display, na->nc, "NickServ", "nickserv/confirm", params, replacements);
```

Indentation here is wrong.

PR #294, opened 2022-04-29
added missing CanSVSO to unreal4 protocol module (tested unreal6)

github-sys-admin:


github-sys-admin:
```
12:23 -!- ***** Svso *****
12:23 -!-  -
12:23 -!-  Gives nick Operflags like the ones in O-Lines.
12:23 -!-  Remember to set SVSMODE +o and alike.
12:23 -!-  Must be sent through an U-Lined server.
12:23 -!-  -
12:23 -!-  Syntax:  SVSO <nick> <+operflags> (Adds the Operflags)
12:23 -!-           SVSO <nick> - (Removes all O-Line flags)
12:23 -!-  Example: SVSO SomeNick +bBkK
12:23 -!-  -
12:23 -!-  -
12:23 -!- UnrealIRCd-6.0.4-git. hub.clandestine.network 0 [*=6000]
12:23 -!- AWAYLEN=256 BOT=B CASEMAPPING=ascii CHANLIMIT=#:256 CHANMODES=beI,fkL,lH,cdimnprstzCDGKMNOPQRSTVZ CHANNELLEN=32 CHANTYPES=# CHATHISTORY=50 
          CLIENTTAGDENY=*,-draft/reply,-draft/typing,-typing DEAF=d ELIST=MNUCT EXCEPTS are supported by this server
12:23 -!- EXTBAN=~,acfjmnpqrtCGOST INVEX KICKLEN=307 KNOCK MAP MAXCHANNELS=256 MAXLIST=b:256,e:256,I:256 MAXNICKLEN=30 MINNICKLEN=2 MODES=12 MONITOR=128 NAMELEN=50 are 
          supported by this server
12:23 -!- NAMESX NETWORK=clandestine NICKLEN=30 PREFIX=(qaohv)~&@%+ QUITLEN=256 SAFELIST SILENCE=64 STATUSMSG=~&@%+ 
TARGMAX=DCCALLOW:16,ISON:,JOIN:16,KICK:16,KILL:16,LIST:16,NAMES:1,NOTICE:16,PART:16,PRIVMSG:16,SAJOIN:16,SAPART:16,TAGMSG:16,USERHOST:,USERIP:,WATCH:,WHOIS:16,WHOWAS:1 
          TOPICLEN=256 UHNAMES USERIP are supported by this server
12:23 -!- WALLCHOPS WATCH=128 WATCHOPTS=A WHOX are supported by this server

```

SadieCat:
Setting that flag does nothing by itself and looking at the Unreal SOURCE it seems like the help entry you have quoted is a remnant left over [from when SVSO was removed](https://github.com/unrealircd/unrealircd/commit/0f2af3f506613cf601341965f04fb388a4f227fd).

github-sys-admin:
I just tested it, the reason you have to set the SVSO class prop is so that it actually enables the module 

```
12:24 -OPERSERV(OS@svc.clandestine.network)-     OLINE          Give Operflags to a certain user
```
works now but its rather useless seeing as how even after doing `oline somenick +o` it's not persistent after somenick reconnects. So yeah that can die for all I care, 


moreover, on the oper module, it's not working right either and Im gonna have to figure out why because I've been trying for like a year to get this to work right and it simply does not: 

```
12:31 <toor> oper list
12:31 -OPERSERV(OS@svc.clandestine.network)- Name     Type
12:31 -OPERSERV(OS@svc.clandestine.network)- toor     clandestine
12:31 -OPERSERV(OS@svc.clandestine.network)-    This oper is configured in the configuration file.
12:31 -OPERSERV(OS@svc.clandestine.network)-    toor is online using this oper block.
12:31 -OPERSERV(OS@svc.clandestine.network)- wtf      clandestine
12:31 -OPERSERV(OS@svc.clandestine.network)-    wtf is online using this oper block.
```

```
 opertype                                                                                                                                                   
 {                                                                                                                                                          
   name                      = "clandestine";                                                                                                               
   commands                  = "*";                                                                                                                         
   privs                     = "*";                                                                                                                         
+  modes                     = "+o"                                                                                                                         
+}                  
```

SadieCat:
>I just tested it, the reason you have to set the SVSO class prop is so that it actually enables the module

The module loading doesn't mean it will actually do anything. The os_oline module only really works with Unreal 3.2-style oper flags so not loading is as intended. In fact, the entire module has been removed in the 2.1 branch along with Unreal 3.2 protocol support.

>moreover, on the oper module, it's not working right either and Im gonna have to figure out why because I've been trying for like a year to get this to work right and it simply does not:

This is unrelated to the os_oline module. If you have any support questions try asking in irc.anope.org #anope.

github-sys-admin:
yeah nevermind I see what your saying *WOOPSIE* sorry 

```
12:38 -!- svso Unknown command
```


ok then maybe the thing to do would be to get rid of this line, 
```
https://github.com/github-sys-admin/anope/blob/2.0/modules/commands/os_oline.cpp#L72
```


To be honest though I'd rather fix the oper module, the oline module is kind of meh

github-sys-admin:
> The module loading doesn't mean it will actually do anything useful. 

I just got done showing you that it works on unreal 6

```
12:35 !svc.clandestine.network Changing your usermodes to +o
12:35 -!- Mode change [+o] for user wtf

```

You just need to remove this from the oline module I'll do this for you in a minute and send you another pull request I'm trying to figure out this other problem right now but I'd appreciate it if you would acknowledge that because I'm kinda feeling a little put off... 

```
                if (!IRCD || !IRCD->CanSVSO)
                        throw ModuleException("Your IRCd does not support OMODE.");
```

SadieCat:
>I just got done showing you that it works on unreal 6

Just because the module did something does not mean that it is what the module is intended to do. The module is intended for setting oper flags on Unreal 3.2 not just generically marking the user as an oper.

>You just need to remove this from the oline module 

Removing this will result in problems on other IRCds which don't allow you to just set `+o` and become an oper.

github-sys-admin:
> doesn't do what it's intended to do 
lol, I mean, you got a point there but I have to ask if not then what 

> Removing this will result in problems on other IRCds 
Yeah I started to look for a way to determine if it's unreal, and that's fair although it's a mod that's disabled; if I can figure out how to check to see if the unreal or unreal4 modules are loaded instead would you settle for that? 

I need to see what oline does that isn't what oper is doing. I understand that lots of ircds like hybrid have gone out of their way to make setting +o any other way than in a config impossible for whatever reason. I don't really care, the fact that I don't have to with unreal is the reason why I'm not using hybrid, plexus, or inspircd. BUT I have setup all of those extensively before arriving at the conclusion each time that none of them were useful to me, and that ultimately the only thing that could be considered satisfactory is being able to set +o without having to put some user's name and some user's password into my config files. I don't even want to give people olines at all but unfortunately it is necessary.

github-sys-admin:
I just fixed this problem by creating my own protocol class and made a patch file that gets applied with docker: 

- https://gitlab.com/allaynawilson/docker-unreal/-/blob/master/anope/0001-added-network-for-clandestine.patch#L809
- https://gitlab.com/allaynawilson/docker-unreal/-/blob/master/anope/0001-added-network-for-clandestine.patch#L809

now I can use umode from my default console user on OPERSERV to set +o to a user and then that user
can make it sticky with keepmodes. 

```
15:22 -!- wtf!toor@100.65.0.9 wtf You are now logged in as wtf.
15:22 !svc.clandestine.network Changing your usermodes to +oW
```

o.o

PR #299, opened 2022-09-15
Update hybrid protocol module to improve compatibility with the lates…

miwob:
…t ircd-hybrid release.

* Remove support for deprecated channel mode `u`
* Add support for channel mode `N`
* Fixed the `UID` message handler requiring the correct amount of arguments
* Fixed the `SID` message handler requiring the correct amount of arguments
* Fixed the `SJOIN` message handler requiring the correct amount of arguments
* Minor documentation updates and style corrections
* Remove pointless return statement in the `JOIN` message handler
* Add support for the owner and admin channel prefix modes
* Add support for `MLOCK`
* Fixed SendChannel() so modes are prefixed with a `+`
* Change SendJoin() to actually use the mode stacker
* Add support for channel mode `Q`
* Add support for channel mode `Z`
* Add support for user mode `B` and `Z`
* Actually user mode `g` is not oper only. For some reason this mode got set to oper only in anope/anope@1e625b6
* Add support for channel mode `V`
* Work properly with extended bans, e.g. don't normalize masks starting with a `$`

### DIFF
--- a/modules/protocol/hybrid.cpp
+++ b/modules/protocol/hybrid.cpp
@@ -1,7 +1,7 @@
 /* ircd-hybrid protocol module. Minimum supported version of ircd-hybrid is 8.2.23.
  *
  * (C) 2003-2022 Anope Team <team@anope.org>
- * (C) 2012-2020 ircd-hybrid development team
+ * (C) 2012-2022 ircd-hybrid development team
  *
  * Please read COPYING and README for further details.
  *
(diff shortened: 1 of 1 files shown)


modules/protocol/hybrid.cpp
```diff
@@ -22,7 +23,7 @@ class HybridProto : public IRCDProto
 		u->KillInternal(source, buf);
 	}
 
-  public:
+ public:
 	HybridProto(Module *creator) : IRCDProto(creator, "ircd-hybrid 8.2.23+")
```
SadieCat, line 27: is hybrid 8.2.23 still the minimum version? If not please update it.

miwob: It's still link compatible to 8.2.23 if that is what you mean. Older hybrid releases just don't have the new user-/channel modes.

PR #302, opened 2022-12-09
Disable `os_session` by default

PeGaSuS-Coder:
**This idea is open to discussion**

Nowadays, IRCds such as UnrealIRCd, InspIRCd, Hybrid IRCd, Solanum, etc, have the ability to define the nnumber of local and global users that can be connected to the network from the same IP as the same time.

IMHO, this module should be disabled by default, since actually we would need to edit 2 configuration files (the ircd and operserv) to allow higher number of users connected from the same IP.

Any ideas/thoghts are welcome.

### DIFF
--- a/data/operserv.example.conf
+++ b/data/operserv.example.conf
@@ -548,8 +548,10 @@ command { service = "OperServ"; name = "RELOAD"; command = "operserv/reload"; pe
  * be killed. Exceptions to the default session limit can be defined via the exception list.
  *
  * Used to manage the session limit exception list, and view currently active sessions.
+ * 
+ * Most of the IRCds nowadays (UnrealIRCd, InspIRCd, Hybrid IRCd, Solanum IRCd) can do this on their own.
  */
-module
+#module
 {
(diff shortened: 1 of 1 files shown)


SadieCat:
I'm neutral on this change. The main benefit of os_session is it allows session limits to be changed without editing config files. OTOH it does cause confusion for a new users when they get killed before hitting their IRCd session limit. Maybe we should at the least raise the default session limit from 3 to 5?

ZarTek-Creole:
I'm in favor of increasing the limit to a reasonable and non-intrusive number. Like a value of 10. People who want to be stricter can lower this value. And those who use robots, services etc whose value 10 is no longer enough will be able to increase. Between two novice users will no longer be bothered by this option.

SadieCat:
I raised the limit from 3 to 5 in commit f0ad0b4f0a153f9ea28d7f34a6f7fd26b0e28cee which should solve the issue that most users have with hitting the session limit. I think that's good enough for now.

PR #308, opened 2023-02-27
fix: make language/update.sh safer, faster, cleaner

henk84:
If it’s too much change for one commit, I could try to split it up. I just thought I’d try it like this first (;


SadieCat:
Merged, apologies for the delay and thanks for your contribution! 👍

PR #310, opened 2023-04-21
normalise realnames in connect logs

examknow:
This prevents un-reset colors in realnames from making the rest of that log line in that color

### DIFF
--- a/src/users.cpp
+++ b/src/users.cpp
@@ -69,7 +69,7 @@ User::User(const Anope::string &snick, const Anope::string &sident, const Anope:
 	{
 		++sserver->users;
 		if (server->IsSynced())
-			Log(this, "connect") << (!vhost.empty() && vhost != host ? "(" + vhost + ") " : "") << "(" << srealname << ") " << (!uip.empty() && uip != host ? "[" + uip + "] " : "") << "connected to the network (" << sserver->GetName() << ")";
+			Log(this, "connect") << (!vhost.empty() && vhost != host ? "(" + vhost + ") " : "") << "(" << Anope::NormaliseBuffer(srealname) << ") " << (!uip.empty() && uip != host ? "[" + uip + "] " : "") << "connected to the network (" << sserver->GetName() << ")";
 	}
 
 	if (UserListByNick.size() > MaxUserCount)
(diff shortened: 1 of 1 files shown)


src/users.cpp
```diff
@@ -69,7 +69,7 @@ User::User(const Anope::string &snick, const Anope::string &sident, const Anope:
 	{
 		++sserver->users;
 		if (server->IsSynced())
-			Log(this, "connect") << (!vhost.empty() && vhost != host ? "(" + vhost + ") " : "") << "(" << srealname << ") " << (!uip.empty() && uip != host ? "[" + uip + "] " : "") << "connected to the network (" << sserver->GetName() << ")";
+			Log(this, "connect") << (!vhost.empty() && vhost != host ? "(" + vhost + ") " : "") << "(" << Anope::NormaliseBuffer(srealname) << ") " << (!uip.empty() && uip != host ? "[" + uip + "] " : "") << "connected to the network (" << sserver->GetName() << ")";
```
SadieCat, line 72:
```suggestion
			Log(this, "connect") << (!vhost.empty() && vhost != host ? "(" + vhost + ") " : "") << "(" << Anope::NormalizeBuffer(srealname) << ") " << (!uip.empty() && uip != host ? "[" + uip + "] " : "") << "connected to the network (" << sserver->GetName() << ")";
```

src/users.cpp
```diff
@@ -295,7 +295,7 @@ User::~User()
 	if (this->server != NULL)
 	{
 		if (this->server->IsSynced())
-			Log(this, "disconnect") << "(" << this->realname << ") disconnected from the network (" << this->server->GetName() << ")";
+			Log(this, "disconnect") << "(" << Anope::NormaliseBuffer(this->realname) << ") disconnected from the network (" << this->server->GetName() << ")";
```
SadieCat, line 298:
```suggestion
			Log(this, "disconnect") << "(" << Anope::NormalizeBuffer(this->realname) << ") disconnected from the network (" << this->server->GetName() << ")";
```

SadieCat:
Unsure if this is the best way to do this. Might be better to just add a reset character in there like @InspIRCd does instead of stripping it.

PR #333, opened 2023-12-04
Adds new Italian translations for 2.1.0

Dragone2:
Hi,
I've updated and added some new translations for the Italian language :)


language/anope.it_IT.po
```diff
@@ -9085,3 +9054,66 @@ msgstr "{nick | canale}"
 
 msgid "{nick | channel} memo-text"
 msgstr "{nick | canale} testo-memo"
+
```
SadieCat, line 9057: Can you delete the old strings from the bottom of the file?

Dragone2: Done :)

PR #334, opened 2023-12-06
messages: fix incorrect format specifier

szsam:
## Summary
Fix incorrect format specifier

## Rationale
Calling a printf-like function with the wrong type of arguments causes unpredictable behavior.

## Testing Environment

**Operating system name and version:** Ubuntu 20.04 
**Compiler name and version:** GCC 9.4.0


SadieCat:
We can't use `%zu` in 2.0 because it requires C++11. Until 2.1 the best solution is to cast to `unsigned long` and use `%lu`.

SadieCat:

Issue #337, opened 2023-12-20
Update the translation files


SadieCat:
Some of the translation files are extremely outdated and we should look into fixing that especially given the new translation strings in 2.1. If you have been mentioned in this issue but don't want to contribute an update then feel free to say that and unsubscribe from the issue.

## Up to date

- [x] en_US (obviously)

## Needs some updates

- [ ] de_DE (688 fuzzy, last updated by @henk84) &mdash; most of these fuzzy strings seem to have been not marked as translated in Poedit during the last update
- [ ] el_GR (679 fuzzy, last updated by @westor7)
- [ ] es_ES (140 fuzzy, no recent translator)
- [ ] fr_FR (52 fuzzy, no recent translator)
- [ ] it_IT (9 fuzzy, last updated by @Dragone2)
- [ ] nl_NL (48 fuzzy, last updated by @Robby-)
- [ ] pl_PL (46 fuzzy, last updated by @k4bek4be)

# Needs significant updates

These are pretty concerning. Given these translations were apparently never updated for 2.0 so the they're barely translated and the translated strings that that exist are quite broken it might be worth deleting if we don't find a maintainer.

- [ ] ca_ES (1028 fuzzy, no recent translator)
- [ ] hu_HU (1028 fuzzy, no recent translator)
- [ ] pt_PT (1028 fuzzy, no recent translator)
- [ ] ru_RU (1028 fuzzy, no recent translator)
- [ ] tr_TR (1028 fuzzy, no recent translator)

SadieCat:
@satmd is currently looking into updating the German translation.

SadieCat:
The Turkish translation has been updated by @CaPaCuL (thanks!).

TehPeGaSuS:
I'm working on the pt_PT translation.
I can't give you an ETA but I'll try to be as quick as possible.

Cheers!

SadieCat:
As of the latest commit:

 - fr_FR &mdash; 67 fuzzy strings
 - it_IT &mdash; 67 fuzzy strings
 - nl_NL &mdash; 67 fuzzy strings
 - pl_PL &mdash; 67 fuzzy strings
 - tr_TR &mdash; 67 fuzzy strings
 - es_ES &mdash; 260 fuzzy strings
 - el_GR &mdash; 751 fuzzy strings
 - de_DE &mdash; 779 fuzzy strings
 - ca_ES &mdash; 1085 fuzzy strings
 - hu_HU &mdash; 1085 fuzzy strings
[...]

PR #339, opened 2023-12-26
Update pl_PL translation for 2.1

k4bek4be:
Update Polish language file as requested in https://github.com/anope/anope/issues/337


SadieCat:
Please remove the obsolete strings from the bottom of the file. There's an option in a menu Poedit to do this if you use that.

PR #342, opened 2024-01-05
fr_FR: Update fuzzy strings

progval:


SadieCat:
Did you run `./language/update.sh` first? I think there should be more than this.

progval:
Done. Most of the changes were English-specific typo-fixes, and replacement with singular they

PR #348, opened 2024-01-23
Added deny_dropping_display option in ns_drop module

H7-25:
## Summary
This option prevents users from deleting their display nick as long as they have aliases in the group.

## Rationale

If you want to deny account name change just disabling  NS SET DISPLAY and NS UNGROUP in not enough, as anyone can drop his display nick ad the next alias available will become the new display nick

## Todo
- Translations
- A modify to OnExpireTick() on nickserv.cpp to let display nick expire only if there's no aliases when this option is enabled

### DIFF
--- a/data/nickserv.example.conf
+++ b/data/nickserv.example.conf
@@ -330,7 +330,16 @@ command { service = "NickServ"; name = "CERT"; command = "nickserv/cert"; }
  *
  * Used for unregistering names.
  */
-module { name = "ns_drop" }
+module { 
+	name = "ns_drop"
+
+	/*
+	 * If set users will have to drop all aliases before can drop their display nick.
+	 * Recommended, if you don't allow your users to change their account name and you have disabled NS SET DISPLAY
+	 * and NS UNGROUP commands.
+	 */	
+	deny_dropping_display = no
+}
 command { service = "NickServ"; name = "DROP"; command = "nickserv/drop"; }
 
 /*
--- a/data/nickserv.example.conf
+++ b/data/nickserv.example.conf
@@ -330,7 +330,8 @@ command { service = "NickServ"; name = "CERT"; command = "nickserv/cert"; }
  *
  * Used for unregistering names.
  */
-module { 
+module
+{ 
 	name = "ns_drop"
 
 	/*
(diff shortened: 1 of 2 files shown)


modules/commands/ns_drop.cpp
```diff
@@ -37,8 +37,17 @@ class CommandNSDrop : public Command
 			return;
 		}
 
+		const NickCore *nc;
+		nc = source.GetAccount();
+
```
SadieCat, line 42:
```suggestion
		const NickCore *nc = source.GetAccount();
```

modules/commands/ns_drop.cpp
```diff
@@ -37,8 +37,17 @@ class CommandNSDrop : public Command
...
 		bool is_mine = source.GetAccount() == na->nc;
 
+		if (Config->GetModule("ns_drop")->Get<bool>("deny_dropping_display") && is_mine && (nc->aliases->size() > 1) && (na->nc->display == nick))
```
SadieCat, line 45:
```suggestion
		if (Config->GetModule(this->module)->Get<bool>("deny_dropping_display") && is_mine && (nc->aliases->size() > 1) && (na->nc->display == nick))
```

modules/commands/ns_drop.cpp
```diff
@@ -62,6 +71,9 @@ class CommandNSDrop : public Command
 				"is dropped you may lose all of your access and channels that\n"
 				"you may own. Any other user will be able to gain control of\n"
 				"this nick."));
+		if (Config->GetModule("ns_drop")->Get<bool>("deny_dropping_display"))
```
SadieCat, line 74:
```suggestion
		if (Config->GetModule(this->module)->Get<bool>("deny_dropping_display"))
```

data/nickserv.example.conf
```diff
@@ -330,7 +330,16 @@ command { service = "NickServ"; name = "CERT"; command = "nickserv/cert"; }
  *
  * Used for unregistering names.
  */
-module { name = "ns_drop" }
+module {
```
SadieCat, line 333:
```suggestion
module
{ 
```

modules/commands/ns_drop.cpp
```diff
@@ -62,6 +71,9 @@ class CommandNSDrop : public Command
...
+			source.Reply(_("You must drop all aliases in the group\n"
+				before you can drop your display nick."));
```
SadieCat, line 76:
```suggestion
			source.Reply(_("You must drop all aliases in the group\n"
				"before you can drop your display nick."));
```

PR #353, opened 2024-02-17
Remove overleft of module with m_ prefix

robert-scheck:
Minor change to remove overleft of module with `m_` prefix, because at least in my build, it's `xmlrpc_main.so` now.


SadieCat:
Good catch!

PR #356, opened 2024-02-21
Update cs_drop.cpp

TehPeGaSuS:
Removed duplicated `with`, which was a typo.

### DIFF
--- a/modules/chanserv/cs_drop.cpp
+++ b/modules/chanserv/cs_drop.cpp
@@ -58,7 +58,7 @@ public:
 				*code = Anope::Random(15);
 			}
 
-			source.Reply(_("Please confirm that you want to drop \002%s\002 with \002DROP %s %s\002."),
+			source.Reply(_("Please confirm that you want to drop \002%s\002 with \002DROP %s %s\002"),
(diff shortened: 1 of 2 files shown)


modules/chanserv/cs_drop.cpp
```diff
@@ -58,7 +58,7 @@ class CommandCSDrop final
 				*code = Anope::Random(15);
 			}
 
-			source.Reply(_("Please confirm that you want to drop \002%s\002 with with \002DROP %s %s\002."),
+			source.Reply(_("Please confirm that you want to drop \002%s\002 with \002DROP %s %s\002."),
```
SadieCat, line 61:
```suggestion
			source.Reply(_("Please confirm that you want to drop \002%s\002 with \002DROP %s %s\002"),
```

We may as well do this at the same time.

SadieCat:
Did this get into any translations yet? I think it should be in en_US?

TehPeGaSuS:
> Did this get into any translations yet? I think it should be in en_US?

I can confirm that it is in the `anope.en_US.po` file

Issue #360, opened 2024-02-26
Add tag support to more protocol modules


sadiepowell:
Anope 2.1 has tag support now but the only protocol module which has support for sending tags (all protocol modules can handle receiving tags) enabled is InspIRCd as thats what I'm most familiar with. We should extend this to other protocol modules but that will require investigation into what servers can accept being sent.

Currently the tags that Anope will send are `+draft/channel-context` (e341cac8d6565044f7390852afb40c5e388121df) and `+draft/reply` (a75afb597b8b4f7e342c393bc3dea868e66f9b4c).

To add support for sending tags a protocol module should implement the `IsTagValid` module like so:

```c++
bool IsTagValid(const Anope::string &tname, const Anope::string &tvalue) override
{
	// ...
}
```

cc: @syzop

syzop:
Sure, I'll have a go, thanks! :)

PR #377, opened 2024-03-09
Enable message-tags support for UnrealIRCd

syzop:
As (sortof) requested in https://github.com/anope/anope/issues/360, this adds message-tags support for the unrealircd protocol module in anope 2.1. Support for this is auto-detected / enabled conditionally when "MTAGS" is present in "PROTOCTL". This is the case in UnrealIRCd 5.0.0 and later.

The change has been tested with current UnrealIRCd 6.1.5-git by using the CS ENTRYMSG functionality where `+draft/channel-context=` correctly shows up on the client-side. And also tested the other way around, with a (modified) UnrealIRCd which did not have "MTAGS", the message tag does not show up there... as expected.

### DIFF
--- a/modules/protocol/unrealircd.cpp
+++ b/modules/protocol/unrealircd.cpp
@@ -444,6 +444,13 @@ private:
 	{
 		Uplink::Send(user, "SVS2MODE", c->name, "-b", u->GetUID());
 	}
+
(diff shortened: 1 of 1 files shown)


SadieCat:
Merged, thanks!

Issue #374, opened 2024-03-06
OS FORBID issue with expiration time and SQLINE


TehPeGaSuS:
**Description**
When using `/OS forbid add chan +5m #testchan Testing` I'm seeing some weird behaviour:

1) The expiry time isn't respected
2) If a user tries to join the forbidden channel, it is added to UnrealIRCd SQLINE (/stats Q) list and it stays there indefinitely, stopping users from joining even if the forbid already expired
3) The only way to remove the channel from UnrealIRCd SQLINE list is to add it with `/os sqline add #testchan Testing` followed by `/os sqline del #testchan`

Now, the main issue is: who's at fault? Anope, UnrealIRCd or both?

**Steps to reproduce the issue:**
_NOTE: This steps were only tested on UnrealIRCd_
1. Add a channel to the forbid list with `/os forbid add chan +5m #testchan Testing`
2. It should expire after 5 minutes, but it doesn't
3. Try to join the forbidden channel. You'll be kicked with the forbid reason and channel will be added to UnrealIRCd SQLINE list and will remain there even if the forbid expires
4. The only way to remove the channel from UnrealIRCd SQLINE list is to use `/os sqline add #testchan Testing` followed by `/os sqline del #testchan`

**Describe the results you received:**
Results are explained on the **Description** bit

**Describe the results you expected:**
1. Expire time should be respected
2. Channel shouldn't be added to UnrealIRCd's SQLINE list
3. If the channel really needs to be added to UnrealIRCd's SQLINE list, then when the forbid expires it should delete the SQLINE

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `services --version`:**
Anope-2.1.4-git (g08a3523) -- build #8, compiled 23:55:43 Mar  4 2024, flags DG

fuadragon:
I confirm issue,  faced the same problem with InspIRCD. 
/OperServ FORBID with +5m time wont expire, untill i do /OperServ FORBID LIST, it expires immediately.
Also /OperServ SQLINE with +5m time wont expire either, only expires with /OperServ SQLINE DEL

Anope-2.1.4-git (g08a3523) services.test.net [77A] InspIRCd 3+ -- (enc_sha256) -- build #1, compiled 17:44:32 Mar  4 2024, flags G

TehPeGaSuS:
Also reported the bug/issue to UnrealIRCd on [https://bugs.unrealircd.org/view.php?id=6390](https://bugs.unrealircd.org/view.php?id=6390)

PR #378, opened 2024-03-09
Fix SQLINE expiry with UnrealIRCd

syzop:
This fixes https://github.com/anope/anope/issues/374 in the sense that we no longer use the `SQLINE` command in server to server traffic to UnrealIRCd when setting a SQLINE, we now use the `TKL` command. This way the expiry time that anope wants can be communicated to UnrealIRCd. Commands like `OS SQLINE ADD +1m ...` work properly now and for `OS FORBID` it will obey the 15 seconds that anope uses (or whatever 'inhabit' is set to by the admin).

This has been tested on UnrealIRCd 6.1.5-git. The `TKL` command syntax should be fine for any UnrealIRCd in the past 20 years or so, so we should be good :D.

### DIFF
--- a/modules/protocol/unrealircd.cpp
+++ b/modules/protocol/unrealircd.cpp
@@ -194,7 +194,7 @@ private:
 	*/
 	void SendSQLine(User *, const XLine *x) override
 	{
-		Uplink::Send("SQLINE", x->mask, x->GetReason());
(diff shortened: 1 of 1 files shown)


SadieCat:
Would it be possible for you to rebase this against the 2.0 branch?

syzop:
Sure, created https://github.com/anope/anope/pull/380 for that.

PR #393, opened 2024-04-08
Rename example.chk to anope.example.chk

TehPeGaSuS:
Rename `example.chk` to `anope.example.chk` for consistency


TehPeGaSuS:
Renamed to `cron.example.sh` as requested.

PR #395, opened 2024-04-13
Fix typo

robert-scheck:
Change is quite obvious, I guess?


ValwareIRC:
change wasn't obvious enough, would you mind please elaboratign

PR #398, opened 2024-04-23
Update nickserv.example.conf

TehPeGaSuS:
## Summary
Change the permission of the `∕NS LIST` command usage so normal users can't see the registered nicknames.

## Rationale
Limit the `∕NS LIST` command to those that have the `nickserv/list` permission, which are usually opers.

IMHO, allowing regular users to see all the registered nicks with the `/NS LIST *` command is a security/privacy issue.

This PR fixes that by restricting the command to those that have the `nickserv/list` permission (usually opers).
## Testing Environment

**Operating system name and version:**  Ubuntu 22.04 LTS
**Compiler name and version:**  gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

### DIFF
--- a/data/anope.example.conf
+++ b/data/anope.example.conf
@@ -766,6 +766,7 @@ log
  *  nickserv/confirm              - Can confirm other users nicknames
  *  nickserv/drop                 - Can drop other users nicks
  *  nickserv/drop/override        - Allows dropping nicks without using a confirmation code
+ *  nickserv/list                 - Can view all registered nickames
  *  nickserv/recover              - Can recover other users nicks
  *  operserv/config               - Can modify services's configuration
  *  operserv/oper/modify          - Can add and remove operators with at most the same privileges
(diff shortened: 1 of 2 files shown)


SadieCat:
Please also document the privilege list in `anope.example.conf`.

SadieCat:
I'm unsure about this. You can opt-out of being seen in this with `/ns SET PRIVATE ON` (or opt-in by having NS_PRIVATE in default)  so I think its fine?

TehPeGaSuS:
Fair enough.

I probably removed the option at some point or something like that.

TehPeGaSuS:
Apparently, at some point, during some Anope testing or something I may have removed the `ns_private` option from `nickserv:module:defaults` and didn't noticed.

Using `/NS SET PRIVATE ON` does hide the registered nicks from the `/NS LIST *` command output, showing only our nickname.

Closing as a WONTFIX

PR #400, opened 2024-05-01
Update chanserv.example.conf

TehPeGaSuS:
## Summary
Add `cs_private` to ChanServ defaults.

## Rationale
This adds the `cs_private` option to ChanServ, which hides channels from the `/CS LIST` command output for non-privileged users, since some channels may be +s (secret).

## Testing Environment

**Operating system name and version:** Ubuntu 22.04.4 LTS
**Compiler name and version:** gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0


SadieCat:
Secret channels will already not be shown in `chanserv/list` unless you have the `chanserv/list` oper privilege so there's no need to hide this.

PR #419, opened 2024-07-03
Add unix socket support to mysql module

ItsOnlyBinary:
## Summary

This pull-request adds the ability to use unix sockets with the mysql module

## Rationale

Unix sockets have less overhead than tcp

## Testing Environment

**Operating system name and version:** ArchLinux 6.9.7-arch1-1 x86_64
**Compiler name and version:** gcc (GCC) 14.1.1 20240522

mariadb 11.4.2-1, mariadb-libs 11.4.2-1 (libmariadbclient)

### DIFF
--- a/modules/extra/mysql.cpp
+++ b/modules/extra/mysql.cpp
@@ -249,7 +249,7 @@ public:
 				const Anope::string &user = block->Get<const Anope::string>("username", "anope");
 				const Anope::string &password = block->Get<const Anope::string>("password");
 				unsigned int port = block->Get<unsigned int>("port", "3306");
-				const Anope::string &socket = block->Get<const Anope::string>("socket", "");
+				const Anope::string &socket = block->Get<const Anope::string>("socket");
 
 				try
 				{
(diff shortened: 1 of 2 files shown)


modules/extra/mysql.cpp
```diff
@@ -248,13 +249,14 @@ class ModuleSQL final
 				const Anope::string &user = block->Get<const Anope::string>("username", "anope");
 				const Anope::string &password = block->Get<const Anope::string>("password");
 				unsigned int port = block->Get<unsigned int>("port", "3306");
+				const Anope::string &socket = block->Get<const Anope::string>("socket", "");
```
SadieCat, line 252:
```suggestion
				const Anope::string &socket = block->Get<const Anope::string>("socket");
```

modules/extra/mysql.cpp
```diff
@@ -496,12 +499,15 @@ void MySQLService::Connect()
 	const unsigned int timeout = 1;
 	mysql_options(this->sql, MYSQL_OPT_CONNECT_TIMEOUT, reinterpret_cast<const char *>(&timeout));
 
-	bool connect = mysql_real_connect(this->sql, this->server.c_str(), this->user.c_str(), this->password.c_str(), this->database.c_str(), this->port, NULL, CLIENT_MULTI_RESULTS);
+	bool connect = mysql_real_connect(this->sql, this->server.c_str(), this->user.c_str(), this->password.c_str(), this->database.c_str(), this->port, this->socket.empty() ? NULL : this->socket.c_str(), CLIENT_MULTI_RESULTS);
```
SadieCat, line 502:
```suggestion
	bool connect = mysql_real_connect(this->sql, this->server.c_str(), this->user.c_str(), this->password.c_str(), this->database.c_str(), this->port, this->socket.empty() ? nullptr : this->socket.c_str(), CLIENT_MULTI_RESULTS);
```

modules/extra/mysql.cpp
```diff
@@ -248,13 +249,14 @@ class ModuleSQL final
...
+				const Anope::string &socket = block->Get<const Anope::string>("socket", "");
 
 				try
 				{
-					auto *ss = new MySQLService(this, connname, database, server, user, password, port);
+					auto *ss = new MySQLService(this, connname, database, server, user, password, port, socket);
 					this->MySQLServices.emplace(connname, ss);
 
-					Log(LOG_NORMAL, "mysql") << "MySQL: Successfully connected to server " << connname << " (" << server << ")";
+					Log(LOG_NORMAL, "mysql") << "MySQL: Successfully connected to server " << connname << " (" << (socket.empty() ? server + ":" + port : socket) << ")";
```
ItsOnlyBinary, line 259: Just to check is this acceptable, or should I have duplicated the line like it did for the debug logging at the end. (also should the debug logging at the end be like this instead?)

SadieCat: Its fine like this.

PR #422, opened 2024-08-06
fix build on some musl systems

wpbirney:
## Summary

Adds `#include <cstring>` to include/services.h as quite a few musl libc systems fail to build without it. It was needed in many places so figured a single add to services.h is best.

## Rationale

To allow musl distros (voidlinux and others) to package without having to patch it

## Testing Environment

locally and on a small network of 50ish users

**Operating system name and version: void linux kernel 6.10
**Compiler name and version: gcc 13.2

### DIFF
--- a/include/services.h
+++ b/include/services.h
@@ -16,6 +16,7 @@
 #include <cstdint>
 #include <cstdio>
 #include <cstdlib>
+#include <cstring>
 
 #include <algorithm>
 #include <bitset>


SadieCat:
What exact error are you getting? We build on Alpine which is also a musl distro and I haven't seen this issue before.

wpbirney:
> What exact error are you getting? We build on Alpine which is also a musl distro and I haven't seen this issue before.

```
[  7%] Building CXX object src/CMakeFiles/anope.dir/config.cpp.o
/home/blackbeard/anope/src/config.cpp: In member function 'Anope::string Configuration::File::Read()':
/home/blackbeard/anope/src/config.cpp:738:28: error: 'strchr' was not declared in this scope
  738 |                 char *nl = strchr(buf, '\n');
      |                            ^~~~~~
/home/blackbeard/anope/src/config.cpp:21:1: note: 'strchr' is defined in header '<cstring>'; did you forget to '#include <cstring>'?
   20 | #include <stack>
  +++ |+#include <cstring>
   21 | #include <stdexcept>
make[2]: *** [src/CMakeFiles/anope.dir/build.make:174: src/CMakeFiles/anope.dir/config.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:612: src/CMakeFiles/anope.dir/all] Error 2
make: *** [Makefile:156: all] Error 2
```

This is the first error, about a dozen more occur in other files if i added <cstring> to config.cpp. all related to strchr, strsignal and various others provided by <cstring>

I cannot reproduce under alpine, but under void-musl it reproduces every time. void-gcc doesn't have this issue

PR #426, opened 2024-08-31
Update anope.example.conf

TehPeGaSuS:
## Summary

Corrected typo on erroneous option value name. `nickserv:registration` only has as options: none, admin, mail. I assume that `mail` is the word that should replace `yes`

## Rationale

There's no `yes` option on `nickserv:registration`


SadieCat:

PR #527, opened 2025-09-05
Correction of #526

crazycatdevs:
Related to #526 : MySQL doesn't accept the timestamp in topictime, must use FROM_UNIXTIME() like it's done in OnTopicUpdated

## Summary

When re-enabling irc2mysql, the table is not filled when chans had a topic: 
`Incorrect datetime value: '1755679391' for column `zeolia_anope`.`chan`.`topictime` at row 1`

## Rationale

This makes the table empty

## Testing Environment

Debian 12, anope 2.0.16

## Checks

Compiled, shutdown anope, restart, disable and re-enable irc2sql

No more errors in log, database filled.

I have ensured that:

- [x] The code I am submitting is my own work and/or I have permission from the author to share it.
- [x] Generative AI (Copilot, ChatGPT, etc) was not used to create any part of this pull request.
- [x] I have documented any features added by this pull request.
- [x] This pull request does not introduce any incompatible API changes (stable branches only, delete if not applicable).


sadiepowell:
Good catch, thanks!

PR #531, opened 2025-10-09
Adding Romanian Language

KidProtect:
## Summary

I made a translation into Romanian language. I am still evaluating it for possible issues, but at first few checks it looks great.

## Rationale

Romania is a country using IRC since 1998, most popular over Undernet. To this day some Romanians still love to use IRC, but unfortunately they can't move on from the environment of 500 levels of access. Since Romanian language was available on that IRCd services from early years, I believe it strongly contributed to that network's popularity. This is why I try to help Anope grow, and because I also never liked Undernet anyway.

## Testing Environment

I tested this on the latest version of Anope from GitHub, in latest InspIRCD.

OS: Debian 12 Bookworm

## Checks

I have ensured that:

- [x] The Romanian translation I am submitting is my own work.
- [x] Generative AI (Copilot, ChatGPT, etc) was NOT used to create any part of this pull request.
- [x] I have adjusted any meanings in Romanian added by this pull request, fully aware of the old latin expression "traduttore traditore".

### DIFF
--- a/language/anope.ro_RO.po
+++ b/language/anope.ro_RO.po
@@ -6340,7 +6340,7 @@ msgstr "Utilizatori (pseudonim): %lu intrări, %lu compartimente, cel mai lung l
 
 #, c-format
 msgid "Users (uid): %lu entries, %lu buckets, longest chain is %zu"
-msgstr ""
+msgstr "Utilizatori (uid): %lu intrări, %lu compartimente, cel mai lung lanț este %zu"
 
 msgid "Users list:"
 msgstr "Listă de utilizatori este după cum urmează:"
--- a/language/anope.ro_RO.po
+++ b/language/anope.ro_RO.po
@@ -1,0 +1,7168 @@
+# Anope IRC Services language file
+# Copyright (C) 2025
+# This file is distributed under the same license as the Anope IRC Services package.
+# Cristian Anghel <kidprotect-cristian@yahoo.com>, 2025.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Anope\n"
+"Report-Msgid-Bugs-To: #anope.ro at TeraNova IRC Network\n"
+"POT-Creation-Date: 2025-08-08 01:05+0100\n"
+"PO-Revision-Date: 2025-10-08 16:50+0100\n"
+"Last-Translator: Cristian Anghel <kidprotect-cristian@yahoo.com>\n"
+"Language-Team: Romanian\n"
+"Language: ro_RO\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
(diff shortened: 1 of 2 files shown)


sadiepowell:
Hello,

Thanks for your contribution. I'll review this in depth when I have time later but for now please can you add `ro_RO.UTF-8` to options:languages (this will allow you to select the language with `/msg NickServ SET LANGUAGE ro_RO.UTF-8`.

Thanks!

KidProtect:
Done. Thanks.

language/anope.ro_RO.po
```diff
@@ -0,0 +1,7168 @@
...
+"Report-Msgid-Bugs-To: #anope.ro at TeraNova IRC Network\n"
+"POT-Creation-Date: 2025-08-08 01:05+0100\n"
+"PO-Revision-Date: 2025-10-08 16:50+0100\n"
+"Last-Translator: Cristian Anghel <kidprotect-cristian@yahoo.com>\n"
+"Language-Team: Romanian\n"
+"Language: ro_RO\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
```
sadiepowell, line 18: Poedit's checker says this the incorrect pluralisation form for Romanian and it should use this instead:

```suggestion
"Plural-Forms: nplurals=3; plural=(n==1 ? 0 : n==0 || (n!=1 && n%100>=1 && n%100<=19) ? 1 : 2);\n"
```

I don't know Romanian so I can't say how correct this is though.

sadiepowell:
There appears to be one untranslated message and two messages which are obsolete (you can run `update.sh` in the language directory to update against the current code) but other than that I'm happy to merge this when you think the translation is complete.

Thanks! :)

language/anope.ro_RO.po
KidProtect: I used that format on purpuse, because special plural did not apply literally to such technical language, so I simply disregarded it completely.

As an example, according to literal Romanian, Anope can not have "Used on: 0 channels" (output example from /bs info Nickserv). It should be without mentioning the digit in Romanian, similar to "no channel" in English; however that would create warning with the missing variable digit. So I kept it simple and technical, using normal plural for 0 values, keeping the digit variables as these are, and resulting in making the .po file having a single plural, without the special literal form for 0 which Poedit refers to as a second plural.

In short, I tried to keep it simple, easier to understand the code and easier for future maintenance as you modify the English file. This is why I defined such simplistic plural system in Poedit, to avoid an extra line - unnecessary in such technical environment.

As a side note: I did not use Poedit to edit the file, I used Geany, because I am more used to see the lines of codes, then used Poedit at the end to see if there are any warnings or if there is something I might have missed.

I updated the file by keeping an eye on your modifications from the en_US file. I have a server with Change Detection software installed, and it notifies me whenever you make a commit that changes the en_US file of Anope. As I was translating in my free time over the past couple of months, I also updated the ro_RO file with your changes from en_US.

KidProtect:
Done.

PR #576, opened 2026-05-20
Updating Romanian Translation

KidProtect:
Updating Romanian Translation


KidProtect:
As previously advised I updated just the Romanian translation file.

PR #580, opened 2026-06-06
Enhance Atheme database import functionality

Kufat:
## Summary

Enhancements:
1. Add ajoin import support. This required a slight refactor of ajoin to expose some structures.
2. Add support for setting founders and advisors for imported channels. This is a bit rough due to how differently Atheme handles founders successors.
3. Ignore certain irrelevant flags when importing. 
4. Add log line for unimplemented akick expiration feature (which I hope to implement.)

Bug fixes:
1. ApplyAccess() was broken for Atheme flags associated with multiple Anope privs.
2. Akicks were only added if the module **wasn't** loaded.
3. Permanent X-lines expired immediately.

## Rationale

Atheme imports were missing important features, had bugs, and were excessively chatty about non-issues.

## Testing Environment

## Checks

I have ensured that:

### DIFF
--- a/data/anope.example.conf
+++ b/data/anope.example.conf
@@ -1086,6 +1086,16 @@ mail
 	database = "atheme.db"
 
 	/*
+	 * Irrelevant Atheme NickServ falgs to silently ignore.
+	 */
+	ns_ignore_flags = "b"
+
+	/*
+	 * Irrelevant Atheme ChanServ falgs to silently ignore.
+	 */
+	cs_ignore_flags = "v"
+
+	/*
 	 * If you have custom data in your Atheme database that you want converted
 	 * to Anope misc data to be shown with cs_set_misc you can configure that
 	 * using one or more cs_set_misc blocks.
--- a/data/anope.example.conf
+++ b/data/anope.example.conf
@@ -1086,16 +1086,6 @@ mail
 	database = "atheme.db"
 
 	/*
-	 * Irrelevant Atheme NickServ falgs to silently ignore.
-	 */
-	ns_ignore_flags = "b"
-
-	/*
-	 * Irrelevant Atheme ChanServ falgs to silently ignore.
-	 */
-	cs_ignore_flags = "v"
-
-	/*
 	 * If you have custom data in your Atheme database that you want converted
 	 * to Anope misc data to be shown with cs_set_misc you can configure that
 	 * using one or more cs_set_misc blocks.
--- a/modules/database/db_atheme.cpp
+++ b/modules/database/db_atheme.cpp
@@ -12,6 +12,7 @@
 //
 // SPDX-License-Identifier: GPL-2.0-only
 
+#include <algorithm>
 
 #include "module.h"
 #include "modules/botserv/badwords.h"
@@ -21,6 +22,7 @@
 #include "modules/chanserv/mode.h"
 #include "modules/hostserv/request.h"
 #include "modules/info.h"
+#include "modules/nickserv/ajoin.h"
 #include "modules/nickserv/cert.h"
 #include "modules/operserv/forbid.h"
 #include "modules/operserv/news.h"
@@ -135,6 +137,7 @@ struct ChannelData final
 
 struct UserData final
 {
+	Anope::string ajoins;
 	Anope::string info_adder;
 	Anope::string info_message;
 	time_t info_ts = 0;
@@ -172,6 +175,8 @@ private:
 	ServiceReference<XLineManager> sqlinemgr;
 	Anope::map<Anope::string> csmiscdata;
 	Anope::map<Anope::string> nsmiscdata;
+	Anope::string csignoreflags;
+	Anope::string nsignoreflags;
 
 	Anope::map<std::function<bool(DBAtheme*,AthemeRow&)>> rowhandlers = {
 		{ "AC",         &DBAtheme::HandleIgnore    },
@@ -236,21 +241,40 @@ private:
 		{ "XL",         &DBAtheme::HandleXL        },
 	};
 
-	void ApplyAccess(Anope::string &in, char flag, Anope::string &out, std::initializer_list<const char*> privs)
+	static bool removeFirstOccurance(Anope::string& in, char c)
 	{
-		for (const auto *priv : privs)
+		auto pos = in.find(c);
+		if (pos != Anope::string::npos)
 		{
-			auto pos = in.find(flag);
-			if (pos != Anope::string::npos)
+			in.erase(pos, 1);
+			return true;
+		}
+		return false;
+	}
+
+	static void removeAll(Anope::string& in, const Anope::string& unwanted)
+	{
+		auto it = std::remove_if(in.begin(), in.end(), [&](char c){
+			return unwanted.find_first_of(c) != unwanted.npos;
+		});
+		in.erase(it, in.end());
+	}
+
+	bool ApplyAccess(Anope::string &in, char flag, Anope::string &out, std::initializer_list<const char*> privs)
+	{
+		const bool flagFound = removeFirstOccurance(in, flag);
+		if (flagFound)
+		{
+			for (const auto *priv : privs)
 			{
 				auto privchar = flags.find(priv);
 				if (privchar != flags.end())
 				{
 					out.push_back(privchar->second);
-					in.erase(pos, 1);
 				}
 			}
 		}
+		return flagFound;
 	}
 
 	void ApplyFlags(Extensible *ext, Anope::string &flags, char flag, const char *extname, bool extend = true)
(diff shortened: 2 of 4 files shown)


data/anope.example.conf
```diff
@@ -1085,6 +1085,16 @@ mail
...
+	/*
+	 * Irrelevant Atheme NickServ falgs to silently ignore.
+	 */
+	ns_ignore_flags = "b"
+
+	/*
+	 * Irrelevant Atheme ChanServ falgs to silently ignore.
+	 */
+	cs_ignore_flags = "v"
+
```
sadiepowell: I'm unsure as to the benefit of having these in the config because the average user won't know what they mean anyway and they'll only ever see the warning once on import.

modules/database/db_atheme.cpp
```diff
@@ -650,6 +673,14 @@ class DBAtheme final
 		ApplyAccess(flags, 't', accessflags, { "TOPIC" });
 		ApplyAccess(flags, 'V', accessflags, { "AUTOVOICE" });
 		ApplyAccess(flags, 'v', accessflags, { "VOICE", "VOICEME" });
+		if (ApplyAccess(flags, 'F', accessflags, { "FOUNDER" }))
+		{
```
sadiepowell: I wonder if there might be a better approach here.

Would it be possible to keep a list of users with the `F` flag in `ChannelData` and then grant the oldest founder and the second oldest successor?

modules/database/db_atheme.cpp
```diff
@@ -1692,6 +1760,60 @@ class DBAtheme final
...
+					{
+						spacesepstream atheme_ajoin_entry(entry);
+						Anope::string channel, key;
+						bool has_key = atheme_ajoin_entry.GetToken(channel);
+						if (has_key)
+						{
+							atheme_ajoin_entry.GetToken(key);
+							Channel *c = Channel::Find(channel);
+							Anope::string k;
+							if (c && c->GetParam("KEY", k) && key != k)
```
sadiepowell, line 1812: I don't think you need to validate this, we trust that the Atheme database isn't corrupt and it the entry gets validated when a user tries to join anyway.

Kufat: I can skip it if you want, but the actual database for my network does have some corruption here. x.x

modules/database/db_atheme.cpp
```diff
@@ -1160,7 +1218,7 @@ class DBAtheme final
 
 		auto *data = userdata.Require(nc);
 		if (key == "private:autojoin")
-			return true; // TODO
+			data->ajoins = value;
```
sadiepowell: Generally where possible the module tries to separate parsing stuff before its put into `UserData` so it might be cleaner if you make ajoins an `Anope::map<Anope::string>` and handled the splitting in HandleMDU something like this (untested):

```c++
irc::commasepstream ajoins(value);
for (Anope::string cname; ajoins.GetToken(cname); )
{
    auto keypos = cname.find(' ');
	if (keypos == Anope::string::npos)
        data->ajoins.emplace(cname, "");
    else
        data->ajoins.emplace(cname.substr(0, keypos), cname.substr(keypos + 1)));
}
````

This would also get rid of your duplicates earlier.

File: data/anope.example.conf (hunk 1 above)
Comment (Kufat) [reply data/anope.example.conf]: OK, will hardcode them.

File: modules/database/db_atheme.cpp (hunk 3 above)
Comment (Kufat) [reply modules/database/db_atheme.cpp:1812]: Tried doing OpserServ UPDATE on Atheme to see if it wrote a clean version; it didn't.

Admittedly a sample size of one is small, but I do think this validation is worth keeping.

File: modules/database/db_atheme.cpp [hunk 5]
```diff
@@ -236,21 +274,40 @@ class DBAtheme final
...
+	static void removeAll(Anope::string& in, const Anope::string& unwanted)
 	{
-		for (const auto *priv : privs)
+			auto it = std::remove_if(in.begin(), in.end(), [&](char c){
+					return unwanted.find_first_of(c) != unwanted.npos;
+			});
+			in.erase(it, in.end());
+	}
+
+	static bool removeFirstOccurance(Anope::string& in, char c)
```
sadiepowell:
```suggestion
	static bool RemoveFirstOccurrence(Anope::string &in, char c)
```

modules/database/db_atheme.cpp
```diff
@@ -236,21 +274,40 @@ class DBAtheme final
 		{ "XL",         &DBAtheme::HandleXL        },
 	};
 
-	void ApplyAccess(Anope::string &in, char flag, Anope::string &out, std::initializer_list<const char*> privs)
+	static void removeAll(Anope::string& in, const Anope::string& unwanted)
```
sadiepowell:
```suggestion
	static void RemoveAll(Anope::string &in, const Anope::string &unwanted)
```

modules/database/db_atheme.cpp
```diff
@@ -1410,6 +1523,7 @@ class DBAtheme final
 		ApplyPassword(nc, flags, pass);
 
 		// No equivalent: bglmNQrS
+		removeFirstOccurance(flags, 'b'); // nick b flag is ephemeral, ignore
```
sadiepowell: What does `b` actually mean here?

modules/database/db_atheme.cpp
```diff
@@ -1149,7 +1239,14 @@ class DBAtheme final
 		auto value = row.GetRemaining();
 
 		if (!row)
+		{
+			if (key == "private:lastquit:message")
```
sadiepowell: Rather than hardcoding this it'd be cleaner to change the method that gets the value to `GetRemaining(bool allow_empty = false)` and then update the check in the method to `if (remaining.empty() && !allow_empty)` and then call `GetRemaining(true)`. 

Its not an error for this to be empty so its fine to just silently ignore it.

modules/database/db_atheme.cpp
```diff
@@ -131,10 +166,13 @@ struct ChannelData final
 	Anope::string suspend_by;
 	Anope::string suspend_reason;
 	time_t suspend_ts = 0;
+	FounderSuccessorCandidate founder_candidate;
+	FounderSuccessorCandidate successor_candidate;
```
sadiepowell: Minor nit: the members of this struct are ordered alphabetically.

modules/database/db_atheme.cpp
```diff
@@ -120,6 +122,39 @@ struct ModeLockData final
...
+
+	bool operator<(const FounderSuccessorCandidate& other)
+	{
+		return std::tie(priority, mtime) < std::tie(other.priority, other.mtime);
+	}
+
+	FounderSuccessorCandidate& operator=(const FounderSuccessorCandidate& other) = default;
+};
+
+const Anope::string FounderSuccessorCandidate::FLAG_PRIORITY = "FSRsaOoHh";
```
sadiepowell, line 156: I'm confused about the amount of flags here. I understand `F` and `S` but the rest don't appear to have anything to do with channel ownership.

From what I understand Atheme should add the founder flag to a new account when an founder expires so it shouldn't be necessary to check other flags here.

modules/database/db_atheme.cpp
Kufat: MU_NOBURSTLOGIN. Basically, if we see this user come in as part of a netburst after a netsplit, don't allow them to be logged in automatically even if they have an account set in the IRCd. 

It's set here when a user registers:

https://github.com/atheme/atheme/blob/b8b23c51cda120f2cfc5af9ac639cd257575f2ce/modules/nickserv/register.c#L153

It's cleared here as part of the code that handles a server disconnecting from the network:

https://github.com/atheme/atheme/blob/b8b23c51cda120f2cfc5af9ac639cd257575f2ce/libathemecore/servers.c#L215

and checked here when a burst login occurs after a server reconnects:

https://github.com/atheme/atheme/blob/b8b23c51cda120f2cfc5af9ac639cd257575f2ce/libathemecore/services.c#L486

I'd expect it to be set again when a user quits normally, or perhaps at login, but it doesn't seem to be.

modules/database/db_atheme.cpp
Kufat: https://github.com/atheme/atheme/blob/master/libathemecore/account.c#L1164 successor determination logic in atheme. What I wrote is broadly similar, although it also counts several other flags and doesn't look at current channel ranks (e.g. +o but no ChanServ entry, that kind of thing.)

Founder _should_ be +F but honestly after poring through my tiny network's DB I'd rather avoid the potential for a channel ending up founderless due to DB corruption. If at least one person has +F, the channel will have a founder with +F. If _nobody_ has +F, I would argue that the logic here does a decent job of picking an arbitrary founder. If you'd rather have the founder slot empty in that case I can make the change.

sadiepowell: The Atheme logic grants +F (CA_FOUNDER) on line 232 and if it can't grant someone founder it drops the channel (line 248-251) so I don't think a founder-less channel can actually exist without being so corrupt that not even Atheme could load the database.

modules/database/db_atheme.cpp
Kufat: Did a quick once-over and I agree that everything else should tolerate an empty `value`, so this should be fine.

PR #581, opened 2026-06-07
Allow specification of swhois priority + use it when displaying NickServ INFO

Kufat:
## Summary
Add support for swhois_ext priority values. They can be manually specified or automatically set based on line order in the config.

## Rationale
For some types of additional profile data, display order can be relevant to user experience.

## Testing Environment

Testing TODO.

~~I have tested this pull request on:~~

## Checks

I have ensured that:

### DIFF
--- a/modules/nickserv/ns_set_misc.cpp
+++ b/modules/nickserv/ns_set_misc.cpp
@@ -329,8 +329,8 @@ public:
 	{
 		command_data.clear();
 		items_by_priority.clear();
-		const auto count = conf.CountBlock("command");
-		for (int i = 0; i < count; ++i)
+
+		for (int i = 0; i < conf.CountBlock("command"); ++i)
 		{
 			const auto &block = conf.GetBlock("command", i);
 			const Anope::string &cmd = block.Get<const Anope::string>("command");
--- a/modules/nickserv/ns_set_misc.cpp
+++ b/modules/nickserv/ns_set_misc.cpp
@@ -26,6 +26,7 @@ struct CommandData final
 	Anope::string pattern;
 	Anope::string syntax;
 	Anope::string title;
+	time_t priority;
 	bool swhois = false;
 };
 
@@ -34,6 +35,7 @@ static Anope::map<CommandData> command_data;
 
 struct NSMiscData;
 static Anope::map<ExtensibleItem<NSMiscData> *> items;
+static std::vector<std::pair<time_t, ExtensibleItem<NSMiscData> *> > priority_ordered_items;
 
 static ExtensibleItem<NSMiscData> *GetItem(const Anope::string &name)
 {
@@ -132,7 +134,7 @@ static void CheckSWhois(User* u, const Anope::string &name, ExtensibleItem<NSMis
 	auto *nc = u->Account();
 	auto *data = nc ? ext->Get(nc) : nullptr;
 	if (data)
-		IRCD->SendSWhois(nickserv, u, name, Anope::Format("%s: %s", GetTitle(ext), data->data.c_str()));
+		IRCD->SendSWhois(nickserv, u, name, it->second.priority, Anope::Format("%s: %s", GetTitle(ext), data->data.c_str()));
 	else
 		IRCD->SendSWhoisDel(nickserv, u, name, "");
 }
@@ -327,7 +329,9 @@ public:
 	void OnReload(Configuration::Conf &conf) override
 	{
 		command_data.clear();
-		for (int i = 0; i < conf.CountBlock("command"); ++i)
+		const auto count = conf.CountBlock("command");
+		priority_ordered_items.resize(count);
+		for (int i = 0; i < count; ++i)
 		{
 			const auto &block = conf.GetBlock("command", i);
 			const Anope::string &cmd = block.Get<const Anope::string>("command");
@@ -341,7 +345,7 @@ public:
 
 			// Force creation of the extension item.
 			const auto extname = GetAttribute(cname);
-			GetItem(extname);
+			auto item = GetItem(extname);
 
 			auto &data = command_data[extname];
 			if (cmd == "nickserv/saset/misc")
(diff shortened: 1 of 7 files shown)


Kufat:
Thinking about typedef'ing `time_t` to something like `swhois_priority` for clarity. (Or just using int since the system clock fallback should never be used and could probably be removed.)

modules/protocol/unrealircd.cpp
```diff
@@ -346,7 +346,7 @@ class UnrealIRCdProto final
 		Uplink::Send("SENDUMODE", 'o', "From " + source.GetName() + ": " + buf);
 	}
 
-	void SendSWhois(const MessageSource &source, User *target, const Anope::string &tag, const Anope::string &message) override
+	void SendSWhois(const MessageSource &source, User *target, const Anope::string &tag, time_t, const Anope::string &message) override
```
sadiepowell:
```suggestion
	void SendSWhois(const MessageSource &source, User *target, const Anope::string &tag, time_t priority, const Anope::string &message) override
```

This should have a name even if its not used.

modules/nickserv/ns_set_misc.cpp
```diff
@@ -26,6 +26,7 @@ struct CommandData final
 	Anope::string pattern;
 	Anope::string syntax;
 	Anope::string title;
+	time_t priority;
```
sadiepowell:
```suggestion
	time_t priority = 0;
```

Not all commands go through the path that initialises this. I don't think its used but its better to be safe.

modules/nickserv/ns_set_misc.cpp
```diff
@@ -354,8 +358,16 @@ class NSSetMisc final
...
+			data.priority = block.Get<time_t>("misc_priority", "0");
+			// If no priority is specified, go by order in the file
+			if (!data.priority)
+			{
+				data.priority = block.GetLineNum() + 1000;
+			}
 			data.swhois = block.Get<bool>("misc_swhois");
+			priority_ordered_items[i] = {data.priority, item};
 		}
+		std::sort(priority_ordered_items.begin(), priority_ordered_items.end());
```
sadiepowell: I wonder if there's a way to avoid having an extra list for this? Possibly by using a map that orders by priority.

sadiepowell:
> Thinking about typedef'ing `time_t` to something like `swhois_priority` for clarity. (Or just using int since the system clock fallback should never be used and could probably be removed.)

Using time_t is probably fine.

modules/nickserv/ns_set_misc.cpp
```diff
@@ -327,7 +329,9 @@ class NSSetMisc final
 	void OnReload(Configuration::Conf &conf) override
 	{
 		command_data.clear();
-		for (int i = 0; i < conf.CountBlock("command"); ++i)
+		const auto count = conf.CountBlock("command");
+		priority_ordered_items.resize(count);
```
sadiepowell: Resizing here isn't correct because only the set commands have info in `priority_ordered_items` resulting in uninitialised reads when its iterated. You should probably clear it and push onto it.

modules/nickserv/ns_set_misc.cpp
Kufat: I thought about that too. Every solution I came up with either needed an extra data structure (of which this was the most efficient, at 16 bytes/element with minimal overhead) or else required more work for each read,

My initial approach did involve adding operator< to CommandData similar to how it's done in swhois_ext but I didn't get anywhere with that.

Kufat:
Tested:
```text
[13:13:41] -NickServ- identifies as:                  male
[13:13:41] -NickServ- uses these pronouns:            he him his
[13:13:41] -NickServ- accepts these pronouns:         they them theirs
[13:13:41] -NickServ- does NOT accept these pronouns: she her hers it its
```

```text
07 Jun 13:13:31 m_spanningtree: S[27] I :00B METADATA 064AAAAAA specialwhois :+ @ns_set_misc:ACCEPTED-PRONOUNS s 5000 :accepts these pronouns: they them theirs
07 Jun 13:13:31 m_spanningtree: S[27] I :00B METADATA 064AAAAAA specialwhois :+ @ns_set_misc:GENDER s 3000 :identifies as: male
07 Jun 13:13:31 m_spanningtree: S[27] I :00B METADATA 064AAAAAA specialwhois :+ @ns_set_misc:PRONOUNS s 4000 :uses these pronouns: he him his
07 Jun 13:13:31 m_spanningtree: S[27] I :00B METADATA 064AAAAAA specialwhois :+ @ns_set_misc:UNACCEPTED-PRONOUNS s 6000 :does NOT accept these pronouns: she her hers it its
```

```text
[13:20:36] Kufat accepts these pronouns: they them theirs
[13:20:36] Kufat identifies as: male
[13:20:36] Kufat uses these pronouns: he him his
[13:20:36] Kufat does NOT accept these pronouns: she her hers it its

[13:24:22] kufat.local NOTE SWHOIS LIST_ENTRY Kufat #1: accepts these pronouns: they them theirs (priority: 5000, flags: oOs)
[13:24:22] kufat.local NOTE SWHOIS LIST_ENTRY Kufat #2: identifies as: male (priority: 3000, flags: oOs)
[13:24:22] kufat.local NOTE SWHOIS LIST_ENTRY Kufat #3: uses these pronouns: he him his (priority: 4000, flags: oOs)
[13:24:22] kufat.local NOTE SWHOIS LIST_ENTRY Kufat #4: does NOT accept these pronouns: she her hers it its (priority: 6000, flags: oOs)
```

Bottom line, this LGTM and is setting priorities correctly but something is still wrong with sorting for display in insp's swhois_ext.

Kufat:
Fixed swhois_ext sorting in https://github.com/inspircd/inspircd-contrib/pull/304/changes and everything now looks good.

modules/nickserv/ns_set_misc.cpp
```diff
@@ -354,8 +359,16 @@ class NSSetMisc final
 			data.pattern = block.Get<const Anope::string>("misc_pattern");
 			data.syntax = block.Get<const Anope::string>("misc_syntax");
 			data.title = block.Get<const Anope::string>("misc_title");
+			data.priority = block.Get<time_t>("misc_priority", "0");
+			// If no priority is specified, go by order in the file
+			if (!data.priority)
+			{
+				data.priority = (default_priority += 1000);
```
sadiepowell: `data.priority = i * 1000` would also work here.

PR #591, opened 2026-07-08
Handle weird corner case where user is 'logged in' as an alias

Kufat:
## Summary

Handle case where, for whatever reason, an IRCd tells us that a user is logged in as a secondary alias rather than the one associated with their NickCore (display name.)

## Rationale
This can come up during slightly wonky database imports.
1. Import database from e.g. Atheme with Anope setting preservedisplay=no
2. User with display name Robert and alias Bob who's currently logged in hasn't used Robert in a year, so Bob becomes the display name
3. Bounce Anope and reimport the database for whatever reason
4. "Bob" isn't a valid display name, since it's back to Robert, so the NickCore isn't found
5. Bob gets Guest-ed.

It's very stupid but it happened.

## Testing Environment

By the time I did my source dive, everyone had re-identified. 
**I have not tested this.** It looks like it should do the thing, though.

Built on:

## Checks

I have ensured that:

### DIFF
--- a/modules/protocol/inspircd.cpp
+++ b/modules/protocol/inspircd.cpp
@@ -1817,6 +1817,18 @@ private:
 			auto *nc = NickCore::Find(value);
 			if (nc)
 				u->Login(nc);
+			else
+			{
+				// Handle corner cases around database rollbacks/inconsistency and users
+				// whose display nick recently expired, causing an alias to become the
+				// display nick.
+				auto *na = NickAlias::Find(value);
+				if (na)
+				{
+					Log() << "User " << u->nick << " is logged in as alias '" << na->nick << "', logging them in as the display nick '" << na->nc->na->nick << "'.";
+					u->Identify(na);
+				}
+			}
 		}
 	}
 
--- a/modules/protocol/inspircd.cpp
+++ b/modules/protocol/inspircd.cpp
@@ -1814,20 +1814,26 @@ private:
 		{
 			// If we're bursting then then the user was probably logged in
 			// during a previous connection.
-			auto *nc = NickCore::Find(value);
-			if (nc)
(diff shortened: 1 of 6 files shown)


sadiepowell:
I'm fine with this in general but I feel it could be done better. I'll think about how and get back to you in the next few days.

sadiepowell:
Okay, so. I'm thinking it should be something like this (using InspIRCd as an example because thats what I'm most familar with):

```c++
static void HandleAccountName(User *u, const Anope::string &value)
{
	// :36D METADATA 9TSAAAAAA accountname
	// :36D METADATA 9TSAAAAAA accountname Sadie
	if (value.empty())
	{
		// The user has been logged out by the IRC server.
		u->Logout();
	}
	else
	{
		// If we're bursting then then the user was probably logged in
		// during a previous connection.
		auto *na = NickAlias::Find(value);
		if (!na)
		{
			// Nick has been dropped, force the IRCd to deauth them.
			IRCD->SendLogout(u);
			return;
		}

		auto *nc = na->nc;
		if (na == nc->na)
		{
			// User is logged into a non-display nick, their display has
			// probably expired due to a config change so reauthenticate
			// them as their new display nick.
			u->Identify(nc->na);
		}
		else
		{
			// User is logged into their display nick.
			u->Login(nc);
		}
	}
}
```

Does this make sense to you?

---

The login vs identify logic seems largely redundant now we've killed nick access lists so I'm thinking about combining the two and making the full identification process a flag.

Kufat:
Should I make the combination change as part of this PR, then?

sadiepowell:
I still need to look more into the implications of merging identify/login so I'll do that after the next release.

sadiepowell:
LGTM, will merge when CI passes for the minor tweak I just made.

Issue #592, opened 2026-07-12
Atheme import sets 'NS set misc' incorrectly


Kufat:
### Description

Atheme import with NS set misc doesn't set the name fully. It seems to work on first run after import but it shows up incorrectly in the DB.
I'll fix this but probably not today, so I'm creating a ticket for it.

### Steps to reproduce the issue

Import DB from atheme with ns_set_misc information
Bounce services

### Describe the results you received

```json
            {
                "data": "male",
                "name": "GENDER",
                "nc": "Kufat"
            },
```

### Describe the results you expected

```json
            {
                "data": "male",
                "name": "ns_set_misc:GENDER",
                "nc": "Kufat"
            },
```

### Anope version

2.1/recent GitHub

sadiepowell:
I don't have the code in front of me but I'm reasonably sure this is how set_misc data is stored internally? I'm not sure why given the prefix is redundant though.

Kufat:
The import from Atheme just had e.g. GENDER without the prefix (first snippet above) and didn't show up in /whois (swhois_ext) or /ns info. When I set it manually with `/ns set GENDER whatever`, the records that were added had the prefix (second snippet above.) When I took services down and did a manual find/replace on the JSON DB to add the prefix and brought services back up, everything worked as expected.

sadiepowell:
Oh I see, I got your snippets the wrong way around.

When I originally wrote the custom import logic I had intended for users to specify the full name when importing but I get how that could be confusing. I'll make it accept both.

Kufat:
Yeah. The reason why it works initially is that nothing cares about the `name` field of the MiscData struct _except_ serialization. So it's fine for the run where the Atheme DB is imported, but the next time services comes up it doesn't find the fields.

I can handle the fix if you like. It'd probably be more memory efficient to just store it without the `ns_set_misc` in memory and only look for/write that prefix during (de)serialization, right?

PR #593, opened 2026-07-14
Fix *_set_misc blocks missing prefix in 'name' and do more thorough config checks

Kufat:
## Summary

The 'name' field of the serializable set misc objects was set to "FOO" instead of "xs_set_misc:FOO", which led to their working fine on initial import but being unreadable after services were restarted.
Also added some more config checking.
## Rationale

Bug fix: obvious

Fatal errors added on the theory that it's better to have the initial import crash with a clear error than to have services come up only to need to be brought down again once the admin reads the log. (Or, worse, to have the config error go unnoticed until a user asks where their data went a week later.)

## Testing Environment

**Operating system name and version:** Ubuntu 24.04.4 / Linux 5.10.102 on WSL
**Compiler name and version:** GCC 13.3.0
insp 4.10.1

Tested with various configuration issues for ns_set_misc. Error messages were emitted and Anope terminated. Tested success case and verified that the ns_set_misc prefix was present in anope.json.

## Checks

I have ensured that:

### DIFF
--- a/modules/database/db_atheme.cpp
+++ b/modules/database/db_atheme.cpp
@@ -1188,9 +1188,8 @@ private:
 			ExtensibleRef<MiscData> extref(name);
 			if (!extref)
 			{
-				Log(this) << "Configuration error: " << Module::name << " cs_set_misc entry " << it->second
-					<< " lacks a corresponding chanserv/set/misc command";
-				throw ConfigException(); // Can't check when reading the config becuase cs_set_misc might not be loaded
+				throw ConfigException(this->name + " cs_set_misc entry " + it->second
+					+ " lacks a corresponding chanserv/set/misc command");
 			}
 
 			auto *data = extref->Set(ci);
@@ -1338,9 +1337,8 @@ private:
 			ExtensibleRef<MiscData> extref(name);
 			if (!extref)
 			{
-				Log(this) << "Configuration error: " << Module::name << " ns_set_misc entry " << it->second
-					<< " lacks a corresponding nickserv/set/misc command";
-				throw ConfigException(); // Can't check when reading the config becuase ns_set_misc might not be loaded
+				throw ConfigException(this->name + " ns_set_misc entry " + it->second
+					+ " lacks a corresponding nickserv/set/misc command");
 			}
 
 			auto *data = extref->Set(nc);
--- a/modules/database/db_atheme.cpp
+++ b/modules/database/db_atheme.cpp
@@ -1160,7 +1160,9 @@ private:
 			return HandleIgnoreMetadata(ci->name, key, value);
 		else if (key.find(':') == Anope::string::npos)
 		{
-			ExtensibleRef<MiscData> extref("cs_set_misc:" + key.upper());
+			// Make a guess that there could be a ChanServ SET MISC entry w/the mystery key's exact (uppercase) name
+			const Anope::string name = "cs_set_misc:" + key.upper();
+			ExtensibleRef<MiscData> extref(name);
 			if (!extref)
 			{
 				Log(this) << "Unknown public channel metadata for " << ci->name << ": " << key << " = " << value;
@@ -1169,11 +1171,12 @@ private:
 
 			auto *data = extref->Set(ci);
 			data->object = ci->name;
-			data->name = key;
+			data->name = name;
 			data->data = value;
 		}
 		else
 		{
+			// Look for a ChanServ SET MISC entry defined in the configuration
 			auto it = csmiscdata.find(key);
 			if (it == csmiscdata.end())
 			{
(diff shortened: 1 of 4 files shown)


Kufat:
Resolves #592

modules/database/db_atheme.cpp
```diff
@@ -1702,7 +1722,17 @@ class DBAtheme final
 		{
 			const auto &anope = data.Get<const Anope::string>("anope");
 			const auto &atheme = data.Get<const Anope::string>("atheme");
-			if (!anope.empty() && !atheme.empty())
+			if (anope.empty())
+			{
+				Log(this) << " missing ns_set_misc anope config entry" << (atheme.empty() ? "" : (" for atheme entry " + atheme));
+				throw new ConfigException;
```
sadiepowell, line 1728: 1. You should just throw the ConfigException rather than a new pointer to it.
2. ConfigException can specify a reason which is written to stdout on start which is better than just logging.
3. If this throws on rehash it will leave the module in a bad state because the maps will be half-initialised. This probably isn't an issue but the usual trick here is to store into a temporary map then swap them once the config has been reloaded.

modules/database/db_atheme.cpp
```diff
@@ -1160,7 +1160,9 @@ class DBAtheme final
 			return HandleIgnoreMetadata(ci->name, key, value);
 		else if (key.find(':') == Anope::string::npos)
 		{
-			ExtensibleRef<MiscData> extref("cs_set_misc:" + key.upper());
+			// Make a guess that there could be a ChanServ SET MISC entry w/the mystery key's exact (uppercase) name
+			const Anope::string name = "cs_set_misc:" + key.upper();
```
sadiepowell, line 1164: Given how much these are being reused now it might be good to add something to the misc header like this to avoid typos.

```c++
#define CHANSERV_SET_MISC_PREFIX "cs_set_misc:"
#define NICKSERV_SET_MISC_PREFIX "ns_set_misc:"
```

modules/database/db_atheme.cpp
```diff
@@ -1314,28 +1321,31 @@ class DBAtheme final
...
-			ExtensibleRef<MiscData> extref("ns_set_misc:" + it->second);
+			const Anope::string name = "ns_set_misc:" + it->second;
+			ExtensibleRef<MiscData> extref(name);
 			if (!extref)
 			{
-				Log(this) << "Unknown imported account metadata for " << nc->display << ": " << key << " = " << value;
-				return true;
+				Log(this) << "Configuration error: " << Module::name << " ns_set_misc entry " << it->second
+					<< " lacks a corresponding nickserv/set/misc command";
+				throw new ConfigException; // Can't check when reading the config becuase ns_set_misc might not be loaded
```
sadiepowell, line 1343: I'm not sure if throwing ConfigException is the right thing to be doing here because it will only get caught by the last-resort exception handler in main(). I'm unsure what might be better though.

modules/database/db_atheme.cpp
Kufat: 1. Not sure why I did it this way; might have been habit from working on legacy crap at my actual job. Fixed
2. I know there was a problem throwing with a reason; I think the issue was that the error didn't make it into the log. Probably because of 1; sigh. Fixed.
3. Addressed

modules/database/db_atheme.cpp
Kufat: I'm unsure too. I _do_ want to kill Anope on failure rather than just erroring and continuing, though. My thinking:
Because this module will (well, should) only be loaded until the import is completed rather than under normal ongoing use, crashing services is ultimately lower-impact than writing an error log line that might not get read + continuing onward with some data missing. Even though it's a regular DB module, it's _used_ in a way that's effectively a one-off utility operation.

Kufat: Re exception handling changes:
```text
[Aug 29 08:52:57 2026] Loading modules...
[Aug 29 08:52:59 2026] HTTPD: Created HTTP server httpd/main
[Aug 29 08:52:59 2026] Using IRCd protocol InspIRCd 4+ (inspircd)
[Aug 29 08:52:59 2026] Loading databases...
[Aug 29 08:52:59 2026] DB_ATHEME: Configuration error: db_atheme ns_set_misc entry PRONOUNS lacks a corresponding nickserv/set/misc command
[Aug 29 08:52:59 2026] Error: Config threw an exception
```
This LGTM now.

sadiepowell: If you put the error in the exception ctor it will show it as the error instead of "config threw an exception".

Kufat:
```
[Aug 29 09:44:56 2026] Loading databases...
[Aug 29 09:44:56 2026] Error: db_atheme ns_set_misc entry PRONOUNS lacks a corresponding nickserv/set/misc command
```

PR #594, opened 2026-07-14
Fix guest nick detection

Kufat:
## Summary

Guest nick routine would return false if not an exact match to 'Guest#####' because the equality checks were still performed after checking for number substitution.
## Rationale

Bugfix
## Testing Environment

Tested this toy version:
```c++
#include <iostream>
#include <string>

inline std::string lower(const std::string &old)
{
	std::string new_string(old);
	for (auto &chr : new_string)
		chr = std::tolower(chr);
	return new_string;
}

bool IsGuestNick(const std::string &nick)
{
	const std::string guestnick = "Guest#####";
	if (guestnick.empty())
		return false; // No guest nick.

	const auto minlen = std::min(nick.length(), guestnick.length());
	for (size_t idx = 0; idx < minlen; ++idx)
	{
	    std::cout << "Index: " << idx << ",";
		if (guestnick[idx] == '#')
		{
		    if (!isdigit(nick[idx]))
			    return false;
		}
		else if (tolower(guestnick[idx]) != tolower(nick[idx]))
			return false;
	}
	return true;
}

int main() {
    std::cout << IsGuestNick("Guest12345");

    return 0;
}
```

## Checks

I have ensured that:

### DIFF
--- a/modules/nickserv/nickserv.cpp
+++ b/modules/nickserv/nickserv.cpp
@@ -201,10 +201,12 @@ public:
 		const auto minlen = std::min(nick.length(), guestnick.length());
 		for (size_t idx = 0; idx < minlen; ++idx)
 		{
-			if (guestnick[idx] == '#' && !isdigit(nick[idx]))
-				return false;
-
-			if (Anope::tolower(guestnick[idx]) != Anope::tolower(nick[idx]))
+			if (guestnick[idx] == '#')
+			{
+				if (!isdigit(nick[idx]))
+					return false;
+			}
+			else if (Anope::tolower(guestnick[idx]) != Anope::tolower(nick[idx]))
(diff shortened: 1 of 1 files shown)


Kufat:
Note that this will return true for the case where e.g. guestnick is `Guest#####` and the user's nickname is `Guest123456`. I don't consider this a bug, as nicknames beginning with a gust nickname shouldn't be registerable (to prevent registration of e.g. `Guest12345_` after a user's client loses connectivity and reconnects before their old nickname pings out. (Not an uncommon case with mobile clients, unfortunately.)

sadiepowell:
The CI build has failed with this but its not your fault, GitHub Actions is just being unreliable as usual. I'll rerun the job later.

PR #595, opened 2026-07-20
Allow verification of registrations + email address changes via single-click links

### LINKED ISSUE #440: Allow people requesting a registered nick to simply click on a link to confirm their email address.
**Description**
Add a URL on WebCPanel that could be called without logging in to confirm an email, passing the passcode and the nick as parameters.

**Why this would be useful**
This would allow people requesting a registered nick to simply click on a link to confirm their email address.

Kufat:
## Summary

Remove the old NickServ confirm module and replace it with one that allows confirmation links to be added to emails and supports confirmation without having to log into the web UI (although it also handles logins correctly.)

Thought about doing password resets as well, but the lack of return codes + fact that IRC login code doesn't log you in on the web meant I would've had to do something really ugly, like looking at the internal structures for the resetpass code or else scraping the command output (which presents i18n problems.)

Also, add functionality to redirect / to the NS info page if a user is already logged in.

This closes #440.

## Rationale

New users are often intimidated by / commands.

## Testing Environment

**Operating system name and version:** Ubuntu 26.04/WSL, Linux kernel 5.10.102.1-microsoft-standard-WSL2
**Compiler name and version:** g++ (Ubuntu 15.2.0-16ubuntu1) 15.2.0

Gave it a few success and failure tests for the registration confirmation path. Didn't test the email change path.

## Checks

I have ensured that:


sadiepowell:
Would it not be better to merge this into the public confirmation page (`modules/webcpanel/pages/confirm.cpp`) rather than the one that requires you to be authenticated (`modules/webcpanel/pages/nickserv/confirm.cpp`)? That one is intended to be linked to from emails so it feels like that'd make more sense.

Kufat:
> Would it not be better to merge this into the public confirmation page (`modules/webcpanel/pages/confirm.cpp`) rather than the one that requires you to be authenticated (`modules/webcpanel/pages/nickserv/confirm.cpp`)? That one is intended to be linked to from emails so it feels like that'd make more sense.

It's a new public (`WebPanelPage` not `WebPanelProtectedPage`) page. I removed the one under /nickserv/confirm.

I used the existing /confirm page when I first started writing this, but the whole thing ended up being a big if block because the two paths shared nothing except the title. Splitting it into a separate page seemed tidier.

PR #596, opened 2026-07-20
Add support for optionally skipping CRs to sendmail (fixes nullmailer)

Kufat:
## Summary
Added a workaround for mailers that don't like CRLF.

## Rationale
Nullmailer was inserting extra lines in my emails. It doesn't need the CR before the LF at the end of each line, and it doesn't handle them properly. 

## Testing Environment

Fired off a quick email with this enabled. Looked fine.

**Operating system name and version:** Ubuntu 26.04/WSL, Linux kernel 5.10.102.1-microsoft-standard-WSL2
**Compiler name and version:** g++ (Ubuntu 15.2.0-16ubuntu1) 15.2.0

## Checks

I have ensured that:

### DIFF
--- a/data/anope.example.conf
+++ b/data/anope.example.conf
@@ -980,7 +980,7 @@ mail
 	/*
 	 * If set, Anope will end lines with LF rather than CRLF when transmitting
 	 * a message to sendmail. This is nonstandard but necessary for nullmailer
-	 * and probably other mailers.
+	 * and probably some other mailers.
 	 */
 	#dontincludecr = yes
 
--- a/data/anope.example.conf
+++ b/data/anope.example.conf
@@ -978,6 +978,13 @@ mail
 	#dontquoteaddresses = yes
 
 	/*
+	 * If set, Anope will end lines with LF rather than CRLF when transmitting
+	 * a message to sendmail. This is nonstandard but necessary for nullmailer
+	 * and probably other mailers.
+	 */
+	#dontincludecr = yes
+
+	/*
 	 * The content type to use when sending emails.
 	 *
 	 * This directive is optional, and is generally only needed if you want to
--- a/src/mail.cpp
+++ b/src/mail.cpp
@@ -51,20 +51,20 @@ void Mail::Message::Run()
 		return;
 	}
 
-	fprintf(pipe, "From: %s%s\n", send_from.c_str(), eol.c_str());
+	fprintf(pipe, "From: %s%s", send_from.c_str(), eol.c_str());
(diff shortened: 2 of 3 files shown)


sadiepowell:
This feels like a bug in Nullmailer that should be fixed upstream imo.

Kufat:
No argument here. I only submitted the PR because the situation was so similar to the existing `dontquoteaddresses` setting.

_Edit:_
Now that I look at it more I'm not sure the CRs should _ever_ be included. It seems like every sendmail implementation including actual sendmail will accept just LF by default and convert to CRLF itself as needed, but not all of them accept CRLF on input. I'm [seeing](https://www.php.net/manual/en/function.mail.php) that qmail has the same behavior as nullmailer.

The claim that this behavior violates the RFC doesn't seem correct to me. The RFC covers the _wire_ format; I don't see a requirement that MTAs accept CRLF-terminated lines.

src/mail.cpp
```diff
@@ -28,6 +28,7 @@ Mail::Message::Message(const Anope::string &sf, const Anope::string &mailto, con
 	, message(m)
 	, content_type(Config->GetBlock("mail").Get<const Anope::string>("content_type", "text/plain; charset=UTF-8"))
 	, dont_quote_addresses(Config->GetBlock("mail").Get<bool>("dontquoteaddresses"))
+	, dont_include_cr(Config->GetBlock("mail").Get<bool>("dontincludecr"))
```
sadiepowell:
```suggestion
	, eol(Config->GetBlock("mail").Get<bool>("dontincludecr") ? "\n" : "\r\n")
```

Then just use `eol.c_str()` later down.

Kufat: Yeah, that's much better. x.x

src/mail.cpp
```diff
@@ -50,20 +51,20 @@ void Mail::Message::Run()
 		return;
 	}
 
-	fprintf(pipe, "From: %s\r\n", send_from.c_str());
```
sadiepowell, line 53: No need to include the \n down here now its been included in `eol`.

data/anope.example.conf
```diff
@@ -977,6 +977,13 @@ mail
 	 */
 	#dontquoteaddresses = yes
 
+	/*
+	 * If set, Anope will end lines with LF rather than CRLF when transmitting
+	 * a message to sendmail. This is nonstandard but necessary for nullmailer
+	 * and probably other mailers.
```
sadiepowell:
```suggestion
	 * and probably some other mailers.
```

PR #601, opened 2026-08-05
2.1+os forbid listmax

TehPeGaSuS:
## Summary

FORBID LIST previously dumped every matching entry unconditionally. With a large forbid list (e.g. importing a public disposable-email blocklist with several thousand entries), sending that many lines in one burst can exceed the uplink's SendQ and disconnect the server.

This adds:
- A listmax config option (default 50, following the same pattern as ns_list/cs_list/hs_list) that caps the default (unranged) FORBID LIST output. listmax = 0 disables the cap.
- An optional from-to range argument (e.g. FORBID LIST EMAIL 101-210) so opers can page through large lists. Since the oper is explicitly choosing the page size here, an explicit range is shown in full and isn't additionally truncated by listmax.
- A summary line reporting both the true total and which entries were actually returned, e.g. End of forbid list - 110/8198 entries shown (101-210).

## Rationale

FORBID LIST previously dumped every matching entry unconditionally. When the forbid list is large (e.g. importing a public disposable-email blocklist with several thousand entries via the os_forbid/file-loading feature), sending that many lines in one burst can exceed the uplink's SendQ and disconnect the server.
This happened when running OS FORBID LIST EMAIL against a ~7k-entry list and it now mirrors the existing listmax convention used elsewhere and adds a range argument so the full list is still reachable without one command blowing the SendQ.
  
## Testing Environment

**Operating system name and version:**  Ubuntu 24.04.3 LTS
**Compiler name and version:**  GCC 14.2.0

## Checks

I have ensured that:

- [x] The code I am submitting is my own work and/or I have permission from the author to share it.
- [ ] Generative AI (Copilot, ChatGPT, etc) was not used to create any part of this pull request.
- [x] I have documented any features added by this pull request (delete if not applicable).
- [x] This pull request does not introduce any incompatible API changes (stable branches only, delete if not applicable).
s pull request changes.

**NOTE:** AI was used but the PR was extensively tested on a live network.


TehPeGaSuS:
I decided to make my own modules to have this behaviour.

